PR #87, opened 2016-04-21
Fix a small error: forcefield.GBVIGenerator has been removed from openmm

yudongqiu:
This should solve the issue when using forcebalance with openmm targets:
/forcebalance/openmmio.py", line 812, in update_simulation
    forcefield.CustomGBGenerator, forcefield.GBSAOBCGenerator, forcefield.GBVIGenerator]
AttributeError: 'module' object has no attribute 'GBVIGenerator'

### DIFF
--- a/src/openmmio.py
+++ b/src/openmmio.py
@@ -36,7 +36,7 @@ except:
     pass
 
 def get_mask(grps):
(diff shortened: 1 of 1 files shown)


leeping:
Thanks!

PR #88, opened 2016-05-28
Shorten AtomName if too long, and free up GPU memory for Interaction_OpenMM targets.

yudongqiu:
Two small changes:

1. When writing PDB files using molecule.py, an error will be raised if the atom name is too long, i.e. Si123. One way to make it work is to shorten it to S123. (First character + Last 3 characters)

2. For Interaction_OpenMM targets, the OpenMM engine will have the OpenMM.simulation object created and left over after the first energy evaluation, causing the use of GPU memory incrementing during the run. One simple way to get around it is deleting the OpenMM.simulation objects after the data is acquired. This will make the optimizer create new simulation objects in each optimization step, causing minimal performance loss.


leeping:
Thanks! Merging. :)

PR #91, opened 2016-07-13
skip rebuilding topology when loading pdb files in interaction_openmm targets, and remove setting 'pmeGridDimensions' manually 

yudongqiu:


leeping:
This all looks good - I think we commented out the default `pmeGridDimensions` for a good reason.  Can you confirm this?

yudongqiu:
Yes. When we were benchmarking the density of water using a larger water box, we found that the hand-picked `pmeGridDimensions` is not sufficient. Commenting out that part allows us to use the default setting determined by OpenMM and results in better agreements with the experiments.

leeping:
Thanks - merged. :) I think this is an important fix.

PR #93, opened 2017-07-26
Merge changes from Yudong, including surface tension target, plotting features and small improvements and bug fixes

yudongqiu:
It's been a long time since our last merge! I will briefly summerize the changes I have made:
1. The surface tension target has been implemented, using a new `nvt.py` script. The openmmio.py and liquid.py is updated with the ability to scale the simulation box, and interface with `nvt.py`. Some improvements has been done to increase performance and decrease memory usage.
2. The `pure_num_grad` mode for liquid.py is implemented by modifying the `submit_jobs()` function and `get()` function. When `pure_num_grad` for a liquid target is set to `True`, `submit_jobs()` will submit 2 additional simulations for each parameter in sub-folders, with the mvals +h and -h. The `get()` function has been split into `get_normal()` and `get_pure_num_grad()`. The `get_normal()` is identical to the original `get()`. The `get_pure_num_grad()` function will call the `get_normal()` to get the property in each sub-folder, then compute the gradients using a 3-point formula. The finalizing step is put into `form_get_result()` to form the objectives.
3. `plot_qm_vs_mm()` functions defined in `abinitio.py` and `interaction.py` will be called if `write_level` is set to >1. The `qm_vs_mm.pdf` plot will help in understanding the fitting results. However when running with `work_queue`, these files will not be transferred back. Therefore we need to run one more FB single-point job locally, with the final parameters, to generate these plots. (Also, in FB parameterization the final parameters are only generated, but haven't been used to calculate the objective functions yet)
4. Some additional changes in `molecule.py`, `nifty.py` are for conveniece purposes and should not affect the original usage.


leeping:
I looked over the whole code and the changes look good.  We should give the user an option to use a different finite difference step size for the pure numerical gradients of thermodynamic properties.  The step size desired by the user may be larger than the usual finite difference step size (which is used for many other things), because the former will have statistical errors in the quantity being differentiated.

We should call it something like `liquid_fdiff_h`, and if the user doesn't set it, then it should default to `finite_difference_h`

yudongqiu:
Based on some benchmarks, the `liquid_fdiff_h` keyword is set to an individual parameter, with the diffault value of 0.01, in the liquid target.

src/abinitio.py
```diff
@@ -799,6 +799,7 @@ def callM(mvals_):
                                           col(M_all_print[:,0])-col(Q_all_print[:,0]),
                                           col(self.boltz_wts)))
             np.savetxt("EnergyCompare.txt", EnergyComparison, header="%11s  %12s  %12s  %12s" % ("QMEnergy", "MMEnergy", "Delta(MM-QM)", "Weight"), fmt="% 12.6e")
+            plot_mm_vs_qm(M_all_print[:,0], Q_all_print[:,0], title='Abinitio '+self.name)
```
yudongqiu: This is called under the condition `if self.writelevel > 0`, so the user can use the keyword `writelevel 1` in the input file to enable it. The performance impact is negligible on my experience.

src/abinitio.py
```diff
@@ -1104,3 +1104,21 @@ def compute_objective_part(SPX,QQ0,Q0,Z,a,n,energy=False,subtract_mean=False,div
         raise RuntimeError('Please pass either 0, 1, 2 to divide')
     return X2
 
+def plot_mm_vs_qm(M, Q, title=''):
+    import matplotlib.pyplot as plt
```
yudongqiu: I think python use a cache for imported modules, so multiple calls of this function will not be affected. https://docs.python.org/2/reference/simple_stmts.html#the-import-statement

src/interaction.py
```diff
@@ -189,9 +191,11 @@ def callM(mvals_, dielectric=False):
         D = emm - self.eqm
         dV = np.zeros((self.FF.np,len(emm)))
 
-        # Dump interaction energies to disk.
-        np.savetxt('M.txt',emm)
-        np.savetxt('Q.txt',self.eqm)
+        if self.writelevel > 0:
```
yudongqiu: I agree. When `writelevel` is specified, the plotting function will depend on matplotlib to function.

src/liquid.py
```diff
@@ -338,10 +363,25 @@ def npt_simulation(self, temperature, pressure, simnum):
                 logger.info("You may tail -f %s/npt.out in another terminal window\n" % os.getcwd())
                 _exec(cmdstr, copy_stderr=True, outfnm='npt.out')
             else:
-                queue_up(wq, command = cmdstr+' &> npt.out',
+                queue_up(wq, command = cmdstr+' > npt.out 2>&1 ',
```
yudongqiu: The old command will crash on Ubuntu 16.04 with newer bash shells. We fixed it by using the new command.

src/liquid.py
```diff
@@ -380,6 +420,7 @@ def print_item(key, heading, physunit):
         print_item("Kappa", "Isothermal Compressibility", "10^-6 bar^-1")
         print_item("Cp", "Isobaric Heat Capacity", "cal mol^-1 K^-1")
         print_item("Eps0", "Dielectric Constant", None)
+        print_item("Surf_ten", "Surface Tension", "mJ m^-2")
```
yudongqiu: I will do so.

leeping:
Thanks!  I think everything looks good.  It would help if you modify the documentation in the appropriate place to mention Matplotlib is an optional dependency; then I'll merge it.

leeping:
Thanks a lot!  Merging.  :)

PR #94, opened 2017-10-06
Avoid installing networkx v2

mohebifar:
Networkx v2 has recently been published, and as a result, the fresh installations of ForceBalance attempt to grab the latest version of networkx which leads to this error:

```
G.nodes[n][name] = values
TypeError: unhashable type: 'dict'
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -112,7 +112,7 @@ def buildKeywordDictionary(args):
     setupKeywords["ext_modules"]       = [CMBAR, DCD, PERMUTE, CONTACT]
     setupKeywords["platforms"]         = ["Linux"]
     setupKeywords["description"]       = "Automated force field optimization."
-    setupKeywords["install_requires"]  = ['networkx>=1.9', 'decorator>=3.4.0']
+    setupKeywords["install_requires"]  = ['networkx>=1.9,<2.0', 'decorator>=3.4.0']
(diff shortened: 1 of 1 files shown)


leeping:
Thank you very much.  Any idea if `n` or `name` is a dict?  We could probably come up with a workaround.

mohebifar:
Actually, `name` is a `dict` here. [molecule.py#L1878](https://github.com/leeping/forcebalance/blob/master/src/molecule.py#L1878)

It seems that it's because the behavior of the `set_node_attributes` on `AttributeError` has changed in [this commit](https://github.com/networkx/networkx/commit/fbc901aa2d30c8dfbe145af747097e92d9b5e5c9#diff-2ae80ac77854385ed27233fcda92e81aR441).

leeping:
It looks like the positions of name and values in `set_node_attributes` were swapped because they made `name` into a keyword argument.  

We could change the ForceBalance code to be consistent and require everyone to update to networkx 2.0.  What do you think?

mohebifar:
I would like to take a look into it and see if there are still other conflicts than the `set_node_attributes` thing by running the tests and trying out the "studies".

leeping:
Sounds great.  Incidentally, I'm about to make ForceBalance Python 3-compatible and that will involve doing much of the same thing.

PR #96, opened 2017-12-20
Added distance and angle functions

jstoppelman:


src/molecule.py
```diff
@@ -1999,6 +1999,26 @@ def measure_dihedrals(self, i, j, k, l):
...
         return phis
+    def measure_distances(self, i, j):
+        dist = []
+        for s in range(self.ns):
+            x1 = self.xyzs[s][i]
+            x2 = self.xyzs[s][j]
+            distance = np.linalg.norm(x1-x2)
+            dist.append(distance)
+        return dist
+    def measure_angles(self, i, j, k):
```
leeping: Can you add a blank line between function definitions?

src/molecule.py
```diff
@@ -1999,6 +1999,26 @@ def measure_dihedrals(self, i, j, k, l):
             #phis.append(phimod)
             #print phimod
         return phis
+    def measure_distances(self, i, j):
```
leeping: Can you add a blank line between function definitions?

PR #98, opened 2018-01-12
Py3compat

leeping:
This PR tracks my progress in making ForceBalance Python 3-compatible.


leeping:
At this point, the test suite runs successfully under Python 2.7 and Python 3.6.  Here is an incomplete list of changes:

- The code that allows for pickling of `ForceField` objects was moved from `nifty.py` (custom Pickler and Unpickler classes) into `forcefield.py` (`__setstate__` and `__getstate__`) methods.
- New `__deepcopy__` method for `Molecule` objects
- `nifty._exec` is fixed up in order to read and write Unicode characters.
- The `pymbar` extension is no longer used, instead using the up-to-date `pymbar` package (https://github.com/choderalab/pymbar)
- Ran `futurize` on each source file; many generators are explicitly cast to lists
- Checked all division operations in each source file

Because test coverage is incomplete, remaining issues in Py3 are still likely.  Leaving this up for a few days for comments before merging.

leeping:
- Wrote a new `AtomContact` function and rewrote `Molecule.write_pdb()` in order to remove all GPL or LGPL code included from other packages.  

The entire package should now be releasable under a more permissive license (e.g. BSD 3-clause).

PR #104, opened 2018-02-22
look for dcdlib in local folder

ChayaSt:
This fix helps `forcebalance` find `dcdlib` when geomeTRIC is called from the command line.


leeping:
Thank you!  I'll merge it now.

PR #105, opened 2018-02-23
More intuitive error message when lxml module is missing

yudongqiu:
Original error "etree not defined" does not indicate the reason, that lxml module is not found. New warning and error messages added to suggest installing the lxml module to solve the problem.


leeping:
Thanks - the failing tests are due to a stochastic failure. Merging.

leeping:
Thanks Yudong - looks like there are several changes in this PR.  I will merge as soon as tests are finished.  

There are some tests with stochastic failures that makes the "X" appear.  Those should be fixed at some point.

PR #110, opened 2018-04-23
Fix bug in liquid.py for Points and mE index mis-allignment, add test_liquid

yudongqiu:
Please see comments on codes for details of bug fix.


src/liquid.py
```diff
@@ -796,8 +797,18 @@ def get_normal(self, mvals, AGrad=True, AHess=True):
         self.AllResults[astrm]['Steps'].append(self.liquid_md_steps)
 
         if len(mPoints) > 0:
-            self.AllResults[astrm]['mE'].append(np.array([i for pt, i in zip(Points,mEnergies) if pt in mPoints]))
```
yudongqiu: The if statement causes a difference in the index of `mE` and `Points`.

leeping:
Could you please check Travis to see why the tests are failing?

leeping:
Thanks, I think it generally looks good.  Can we change `Reference` back to `CUDA` in the `AbInitio_OpenMM` and `Hydration_OpenMM` targets?  Also, is it really necessary to set `OMP_NUM_THREADS=1`?  Can we unset that environment variable?

yudongqiu:
Thanks for the quick feedback. Sure I will change the default platform from Reference back to CUDA for these targets. I think setting OMP_NUM_THREADS for tests to be 1 is consistent with the configure for calling `gmx mdrun -t 1` but unset it will allow other programs to use more than one core by default if needed.

yudongqiu:
I think now the tests in Python2 works but not Python3. The reason is because the incompatibility of `npt_result.p` created by lp_dump() in Python2 and the lp_load() in Python 3 fails to load it. Is there a good way to solve this?

leeping:
We could skip over the test in one of the major versions, what do you think?  I'd prefer skipping over Python 2 because that's the version with a shorter lifespan.

yudongqiu:
Sure I have converted all the .p files to python 3.

PR #111, opened 2018-05-02
add tools/install-cctools-62.sh that works for both Python 2 and 3

yudongqiu:
This script will be useful when installing a fully functional ForceBalance package.
The previous install script only works for Python 2.
Since the problem of installing work_queue Python 3 has been solved, the new script will now automatically detect which version of python is in use, configure and install it correctly.


yudongqiu:
Closed this because Lee-Ping happened to write the same script at the same time by coincidence.

PR #116, opened 2018-09-10
Update WQ printout to use wq_stats from CCTools 6.2.10

leeping:


leeping:
The tests work locally, so I should figure out what's wrong with the tests on Travis.

PR #117, opened 2018-10-14
Mix xml txt

leeping:
The previous forcefield.py would crash if we mixed XML and text parameter files in a single optimization if both files contained free parameters. This PR fixes the issue.


leeping:
All tests passed locally.

PR #133, opened 2019-01-29
H-Bond restrains for OpenMM and SMIRNOFF

MSchauperl:
OpenMM h-bond restraints are accessible via input file argument (restrain_h). For hydrogen bond restraints with SMIRNOFF an example SMIRNOFF FF file was created in the SMIRNOFF Example folder.


leeping:
I changed the word "restrain" to "constrain" in a few places.

MSchauperl:
I changed the optimize.out file back to the original

PR #134, opened 2019-02-04
Repair constrain

MSchauperl:
We forgot to rename restrain_h to constrain_h in openmmio.py


leeping:
Thanks - it turns out I had made the same change a while ago.  The PR is merged now.  :)

PR #135, opened 2019-02-13
OpenMM Normal Modes

yudongqiu:
The normal mode analysis for OpenMM engine is implemented. 

I chose a finite difference of 0.001 Angstrom for computing the Hessian matrix after some testing. 

However, it is still not possible to match the normal modes to within 0.01 of the Tinker value. 

I guess it might be caused by errors introduced by finite difference. 

Widen the tolerance to 0.02 allows the tests to pass.

Edit 03.19.2019:

Code added for enabling "Vibration_OpenMM" target in command line.

One bug fixed with `self.FF.restrain_h`

`studies/020_openmm_normal_mode` added, with a few example targets copied from `studies/015_amoeba_tinker`. One file `Monomer_Vibration/input.pdb` is created by hand to enable the fitting of vibrational data using OpenMM Engine.


src/openmmio.py
```diff
@@ -669,7 +669,7 @@ def prepare(self, pbc=False, mmopts={}, **kwargs):
                 elif self.FF.amoeba_pol == 'direct':
                     self.mmopts['polarization'] = 'direct'
             self.mmopts['rigidWater'] = self.FF.rigid_water
-            if self.FF.restrain_h== True:
+            if self.FF.constrain_h is True:
```
yudongqiu: I think this is a typo. I found it because it raises an `AttributeError` in examples with OpenMM engine.

leeping:
Thanks a lot, LGTM!

PR #144, opened 2019-06-27
improve stability of hessian diagonalization

yudongqiu:
This could help avoid this error:
```
TypeError: Cannot cast ufunc add output from dtype('complex128') to dtype('float64') with casting rule 'same_kind'
```


leeping:
I've never seen this before, is this for a new target or an existing one? Are you sure the Hessian is being computed correctly?

PR #148, opened 2019-07-08
Use the OpenFF API for force field modification

leeping:
This PR uses the OpenFF API to apply changes to parameters in the SMIRNOFF (and all OpenFF-style) force fields.  This should work with all parameter types, "free" as well as "evaluated" parameters, and all targets.

One "todo" is to identify which hacks in `smirnoff_hack.py` are no longer necessary; @yudongqiu might want to take a look. Also tagging @j-wags .

### DIFF
--- a/src/smirnoff_hack.py
+++ b/src/smirnoff_hack.py
@@ -49,7 +49,7 @@ OpenEyeToolkitWrapper.generate_conformers = cached_generate_conformers
 
 # final timing: 56s
 
-# cache the ForceField creation
+# cache the ForceField creation (no longer needed since using OpenFF API for parameter modifications)
 
 import hashlib
 from openforcefield.typing.engines.smirnoff import ForceField
--- a/src/smirnoffio.py
+++ b/src/smirnoffio.py
@@ -138,6 +138,44 @@ class SMIRNOFF_Reader(BaseReader):
             logger.info("Minor warning: Parameter ID %s doesn't contain any SMIRKS patterns, redundancies are possible\n" % ("/".join([InteractionType, parameter])))
             return "/".join([ParentType, InteractionType, parameter])
 
+def assign_openff_parameter(ff, new_value, pid):
+    """
+    Assign a SMIRNOFF parameter given the openforcefield.ForceField object, the desired parameter value,
+    and the parameter's unique ID.
+    """
+    # Split the parameter's unique ID into four fields using a slash:
+    # Input: ProperTorsions/Proper/k1/[*:1]~[#6X3:2]:[#6X3:3]~[*:4]
+    # Output: ProperTorsions, Proper, k1, [*:1]~[#6X3:2]:[#6X3:3]~[*:4]
+    # The first, third and fourth fields will be used for parameter assignment.
+    # We use "value_name" to describe names of individual numerical values within a single parameter type
+    # e.g. k1 in the above example.
+    (handler_name, tag_name, value_name, smirks) = pid.split('/')
+
+    # Get the OpenFF parameter object
+    parameter = ff.get_parameter_handler(handler_name).parameters[smirks]
+
+    if hasattr(parameter, value_name):
+        # If the value name is an attribute of the parameter then we set it directly.
+        unit = getattr(parameter, value_name).unit
+        setattr(parameter, value_name, new_value*unit)
+    else:
+        # If the value name is a periodic attribute (say, k1) then we need to use
+        # a regex to split the value name into 'k' and '1', then set the appropriate
+        # value in the k-list
+        attribute_split = re.split(r'(\d+)', value_name)
+        # print(attribute_split)
+        # assert len(attribute_split) == 2
+        assert hasattr(parameter, attribute_split[0])
+        # attribute_split[0] is a string such as 'k'
+        value_name = attribute_split[0]
+        # parameter_index is the position of k1 in the values associated with 'k'
+        parameter_index = int(attribute_split[1]) - 1
+        # Get the list of values, update the appropriate one and then set the new attribute to the updated list
+        value_list = getattr(parameter, value_name)
+        unit = value_list[parameter_index].unit
+        value_list[parameter_index] = new_value*unit
+        setattr(parameter, value_name, value_list)
+            
 class SMIRNOFF(OpenMM):
 
     """ Derived from Engine object for carrying out OpenMM calculations that use the SMIRNOFF force field. """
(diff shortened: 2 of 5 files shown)


yudongqiu:
`smirnoff_hack.getForceField` is used to cache the building of the ForceField object. If this is not needed we can remove this and switch back to use the original initializer.

src/smirnoffio.py
```diff
@@ -138,6 +138,44 @@ def build_pid(self, element, parameter):
...
+    # e.g. k1 in the above example.
+    (handler_name, tag_name, value_name, smirks) = pid.split('/')
+
+    # Get the OpenFF parameter object
+    parameter = ff.get_parameter_handler(handler_name).parameters[smirks]
+
+    if hasattr(parameter, value_name):
+        # If the value name is an attribute of the parameter then we set it directly.
+        unit = getattr(parameter, value_name).unit
+        setattr(parameter, value_name, new_value*unit)
```
yudongqiu: I think this is the "API" call for changing the parameter values, consistent with the [example](https://github.com/openforcefield/openforcefield/blob/master/examples/forcefield_modification/ManipulateParameters.ipynb)
However this is not exactly "clean", as `value_name` can be other things that mistakenly override an unexpected attribute of this object. Also, I think the "k1", "k2" are intrinsic parameter names, so `parameter.k = [k1, k2]` may not be "safe" (i.e. if in the future there is a k0 before k1, or there is k1,k3 but skipped k2).
All of the above potential issues can be solve by an improved API in OpenFF toolkit, for example:
```
> parameter.get_name_values()
{
    'k1': 1.0,
    'periodicity1': 1.0,
    'k2': 3.0,
}

> parameter.get_value('k1')
1.0

> parameter.set_value('k1', 2.0)
> parameter.get_value('k1')
2.0

> parameter.set_value('k_random', 2.0)
ValueError: value with name "k_random" does not exist.
```
@j-wags

leeping:
This pull request **introduces 1 alert** and **fixes 1** when merging 1af89c706407b4a99e449b26acbc87adb088a3d3 into cd9e1a2e6fcc90590ae094e873fa7be35529287c - [view on LGTM.com](https://lgtm.com/projects/g/leeping/forcebalance/rev/pr-e1714891745e0eab8fa9c6561347f7395ca6ba39)

**new alerts:**

* 1 for Unused import

**fixed alerts:**

* 1 for Unused import

---

---

leeping:
@yudongqiu Can you please confirm that all of the unnecessary calls to `smirnoff_hack.getForceField` have been removed from the latest commit?

yudongqiu:
> @yudongqiu Can you please confirm that all of the unnecessary calls to `smirnoff_hack.getForceField` have been removed from the latest commit?

Yes they are all removed. We can remove this function from smirnoff_hack.py too.

src/smirnoffio.py
j-wags: > However this is not exactly "clean", as value_name can be other things that mistakenly override an unexpected attribute of this object.

Could you elaborate? We're trying to enumerate the scope of allowed parameter names so we don't run into trouble with XML/Python/other technology variable name restrictions. This could be useful input for the discussion here -- "Which attribute names should we allow?" https://github.com/openforcefield/openforcefield/issues/338 

> Also, I think the "k1", "k2" are intrinsic parameter names, so `parameter.k = [k1, k2]` may not be "safe" (i.e. if in the future there is a k0 before k1, or there is k1,k3 but skipped k2).

In the SMIRNOFF spec, we've been thinking about how to avoid any sort of misinterpretation along these lines. Right now, [indexed attributes must start at 1, and be consecutive](https://open-forcefield-toolkit.readthedocs.io/en/latest/smirnoff.html#propertorsions). There must also be an equal number of `k#`, `periodicity#` and `phase#` values in torsions (`idivf#` is a different beast, but that's just going to be tricky no matter what we do). 

> `> parameter.get_name_values()`

That's a good suggestion. Does the [`parameter.to_dict()` function](https://open-forcefield-toolkit.readthedocs.io/en/latest/api/generated/openforcefield.typing.engines.smirnoff.parameters.ParameterType.html#openforcefield.typing.engines.smirnoff.parameters.ParameterType.to_dict) come close to satisfying this need? If not, we can consider other options. 

> `> parameter.get_value('k1')`

Hopefully we can access for that in, using the `getattr`/`setattr` functions. I'll try to review [Andrea's PR](https://github.com/openforcefield/openforcefield/pull/377) on that ASAP

src/smirnoffio.py
```diff
@@ -138,6 +137,44 @@ def build_pid(self, element, parameter):
...
+    else:
+        # If the value name is a periodic attribute (say, k1) then we need to use
+        # a regex to split the value name into 'k' and '1', then set the appropriate
+        # value in the k-list
+        attribute_split = re.split(r'(\d+)', value_name)
```
j-wags: I can't believe we didn't expose access to this in the original API. Thanks for being the crash test dummies on this interface, and sorry that it's coming up at such a critical phase!

src/smirnoff_hack.py
```diff
@@ -62,4 +62,4 @@ def getForceField(*ffpaths):
     cache_key = hasher.hexdigest()
     if cache_key not in SMIRNOFF_FORCE_FIELD_CACHE:
         SMIRNOFF_FORCE_FIELD_CACHE[cache_key] = ForceField(*ffpaths, allow_cosmetic_attributes=True)
```
j-wags: This is really clever. Perhaps related -- if it becomes necessary to _compare_ force fields in the future, I think we could do it using something like a hash of the output of `ff.to_string('XML')`. That should standardize whitespace and _I think_ that recent changes may ensure that parameters are written out in a consistent order. Floating point values may still be trouble, but, if this sort of thing ever became important, I could put some time into ensuring that comparison works.

leeping: I think a good discussion on how to compare force fields is well worth it. I'd like to see us eventually have the ability to apply a parameter change to a system, and compute energies at a fraction of the cost of recomputing the energy of the whole system.

j-wags: Agreed -- I think @simonboothroyd is also looking for this functionality. Hopefully @andrearizzi's work on a dedicated OpenFF `System` class could get us closer to this as well.

src/smirnoffio.py
j-wags: The question of how to _add new torsion terms_ still remains. I've opened a spot to discuss this: https://github.com/openforcefield/openforcefield/issues/378

In the current implementation, I think adding new terms would require extending the list of `k` values (along with `periodicity` and `phase`). But we probably want to have a friendlier option, since the interdependence between the separate lists is quite important.

yudongqiu: Sure I'm glad to hear that these improvements are on the schedule.

leeping:
Thanks everyone! I'm merging it now.

PR #149, opened 2019-07-09
improve assign_openff_parameter performance by caching parameter search

yudongqiu:
The purpose of this PR is to improve the performance of the `smirnoffio. assign_openff_parameter()` function.

### Background
This function is introduced in recent PR #148. In theory, updating parameters the OpenFF ForceField object through the API should be faster, I found in my benchmark that it actually become a bit slower than the previous version that cached the ForceField initialization. (Still faster than the original version without cache).
The reason for the performance regression is that the `assign_openff_parameter()` function is called `n_param` times each time the ForceBalance forcefield is updated in `make()`, and numerical displacements require make() function to be called `2 x n_param` times. Therefore the total scaling of the time that the `assign_openff_parameter()` is called is `2 x n_param^2`. This become costly in my benchmark with 508 parameters.

### Method
In the PR we simply cache the parameter finding in `assign_openff_parameter()`, so this function become basically "free" after the initial parameter finding. This effectively reduce the total cost to `n_param`.

### Timing
Here are some timing numbers in my benchmark of one target evaluation.
```
                        [original]    [cached_FF_init]    [modify_api]   [cached_modify_api]
[update forcefield]      41.8             19.9              26.2             0.26
[create system]          10.7             10.8              10.5             10.5
[wall]                   78.7             56.1              62.4             35.5
```
The [create system] is the time cost of `create_openmm_system` after smirnoff_hack, shown here as a reference timing between runs.

### Note
Further performance optimization can be done by only call assign_openff_parameter() function for updated parameters. However since the cached `assign_openff_parameter()` is very fast, the further optimization is less useful.

### DIFF
--- a/src/parser.py
+++ b/src/parser.py
@@ -322,7 +322,7 @@ def read_mvals(fobj):
     for line in fobj:
         if re.match("(/read_mvals)|(^\$end)",line):
             break
-        Answer.append(float(line.split('[')[-1].split(']')[0].split()[-1]))
+        Answer.append(float(line.split('[', maxsplit=1)[-1].split(']', maxsplit=1)[0].split()[-1]))
     return Answer
 
 def read_pvals(fobj):
@@ -330,7 +330,7 @@ def read_pvals(fobj):
     for line in fobj:
         if re.match("(/read_pvals)|(^\$end)",line):
             break
-        Answer.append(float(line.split('[')[-1].split(']')[0].split()[-1]))
+        Answer.append(float(line.split('[', maxsplit=1)[-1].split(']', maxsplit=1)[0].split()[-1]))
     return Answer
 
 def read_priors(fobj):
--- a/src/smirnoffio.py
+++ b/src/smirnoffio.py
@@ -148,33 +148,44 @@ def assign_openff_parameter(ff, new_value, pid):
     # The first, third and fourth fields will be used for parameter assignment.
     # We use "value_name" to describe names of individual numerical values within a single parameter type
     # e.g. k1 in the above example.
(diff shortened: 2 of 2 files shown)


leeping:
This pull request **introduces 1 alert** when merging 18c304292541992a6dcdf9f1ef65e5eb6a624765 into 91e3be696d91489d289ca13444435e9d6180b520 - [view on LGTM.com](https://lgtm.com/projects/g/leeping/forcebalance/rev/pr-eb5d55240ed07609039d6946ee42cc395597d559)

**new alerts:**

* 1 for Unused local variable

---

---

leeping:
I think this is a very nice improvement and it should be merged. @j-wags , any thoughts?

leeping:
This pull request **introduces 1 alert** when merging 9ee32634202fbfe017e47c45f89572a9c5fca548 into 91e3be696d91489d289ca13444435e9d6180b520 - [view on LGTM.com](https://lgtm.com/projects/g/leeping/forcebalance/rev/pr-cdc351decc34ef709407449900d8d37ce0ce5378)

**new alerts:**

* 1 for Unused local variable

---

---

j-wags:
Ah! I'm really sorry, I just saw this. Please feel free to DM me if I leave you hanging like this again.

This optimization is a great and timely solution. I'm opening an Issue on the toolkit repo (https://github.com/openforcefield/openforcefield/issues/383) to improve the performance of these methods, so hopefully we can revisit this in the future. As it stands, if we change the internal structure of our FF object, we risk breaking FB. But I definitely agree that this performance optimization is critical to our initial fitting, and adding the optimization to the toolkit would have taken longer. Thanks for your quick action!

PR #150, opened 2019-07-17
Improve optgeo target

yudongqiu:
This PR improves the `opt_geo_target.py` in 3 ways.

### 1. The gradients are improved. 

Previously the `v_obj_list` contains objective values from each system and each type of internal coordinates: Bond, Angle, Dihedral, Improper. Because individual objective term such as `diff_bond1 = MM_bond1 - QM_bond1` are combined into the group, the gradients are computed as `d(RMSD of all bonds for a system) / d(parameter)`. 

This update further expands the `v_obj_list`, to directly include all individual terms, thus the gradient is calculated more directly as `d(diff_bond1) / d(parameter)`

Note: This change does not affect the objective values.

### 2. The denominator `1/sqrt(n_bonds)` is removed.

Previously when all Bonds are combined into one group for computing RMSD, the denominator contains the term `1/sqrt(n_bonds)`. This denominator is helpful to balance the impact of increasing number of internal coordinates. For example, a small molecule has only 2 bonds, and a large molecule may have 100 bonds. This denominator will greatly scale down contributions from each bond in the large molecule, so the total contribution to objective function is "even" between the two molecules.

The reason we want to remove the denominator here, is that we prefer a more even contribution from each individual internal coordinate, over the "fairness" between molecules. The large molecule contain more bonds should be more important in the fitting. If a large molecule have only one important internal coordinate (IC) among many, removing this denominator will ensure that important one to have an even contribution compared to other internal coordinates in smaller molecules.

Note: This change will increase the total objective value `X2` by `n_IC` times, in practice the increase is about 20x, so a lower weight for the target might be needed.

### 3. The units for all internal coordinate angles are changed from radian to degree.

As suggested by @leeping, it is easier to read and understand in units of degree. This means all denominators defined in the `opt_geo_target.txt` file will have a different unit. And we will be also be printing the values in the new unit in both the output file and the rmsd_decomposition.txt files.

Note: This will cause all previous input files to be invalid. The `angle_denom`, `torsion_denom` and `improper_denom` values needs to be changed to the new unit.

## Others

There is also two small improvements in `nifty.py` related  work_queue.
1. Allow the work_queue to use the fastest worker available
2. Prevent the workers from "freezing".  The `WORK_QUEUE.specify_keepalive_interval(8640000)` command sets a really high interval for host to check if worker is alive. This means if a worker suddenly disconnect without letting the host know, the host will assume the worker is still running a task and wait forever.
    I have encountered this a few times and have to kill and restart the fitting because of this, so I think it's better to comment this out.


leeping:
Thanks; I think this looks good!

yudongqiu:
After a few test fitting runs I think it works as expected. Merging now.

PR #153, opened 2019-08-03
Created torsion profile target

leeping:
This PR adds a new target `TorsionProfile` which is implemented for OpenMM and SMIRNOFF engines.  The fitting data consists of a QM torsion drive and its metadata.  The required files are the same as the `abinitio` target, except `metadata.json` from QCArchive is required.

Prior to fitting the energies, the MM structure is relaxed with a harmonic energy restraint with a default force constant of 1.0 kcal/mol/Å (user adjustable). The four atoms in the torsion are frozen. The main idea is that the we want to match energies at the MM minima and QM minima, since we cannot get MM minima and QM minima to be exactly the same.

Additionally, the `Engine.optimize` API has been modified to always require the snapshot number, and it returns the optimized geometry as a `forcebalance.Molecule` object.


src/openmmio.py
```diff
@@ -888,7 +919,27 @@ def update_simulation(self, **kwargs):
...
+        """
+        Set reference positions for energy restraints.  This may be a different set of positions
+        from the "current" positions that are stored in self.mol and self.xyz_omm.
+        """
+        if hasattr(self, 'restrain_k'):
+            ## Generate OpenMM-compatible positions in nanometers
+            xyz = self.ref_mol.xyzs[shot] / 10.0
+            have_restraint_frc = False
+            for frc in self.simulation.system.getForces():
+                if isinstance(frc, CustomExternalForce):
```
yudongqiu: I think this may have a potential risk to mix up with a OpenMM system that has an CustomExternalForce defined for other purpose. I think a safer and faster way to access the force is to save the index when the force is added to the system. For example in line 528 above:
```
self.restraint_frc_index = self.system.addForce(restraint_frc)
```
then here we can do
```
frc = self.simulation.system.getForce(self.restraint_frc_index)
```

src/openmmio.py
```diff
@@ -888,7 +919,27 @@ def update_simulation(self, **kwargs):
...
+                    have_restraint_frc = True
+                    for i, j in enumerate(self.realAtomIdxs):
+                        frc.setParticleParameters(i, j, [xyz[i, 0], xyz[i, 1], xyz[i, 2]])
```
yudongqiu: This is a small suggestion: I wonder if `frc.setParticleParameters(i, j, xyz[i])` works here, if not we can also consider replacing line 929 with
```
xyz = (self.ref_mol.xyzs[shot] / 10.0).tolist()
```

src/openmmio.py
```diff
@@ -1117,34 +1168,68 @@ def normal_modes(self, shot=0, optimize=True):
...
+        # printcool_dictionary(energy_components(self.simulation), title='Energy component analysis before minimization, shot %i' % shot)
         # Minimize the energy.  Optimizer works best in "steps".
         for logc in np.linspace(0, np.log10(crit), steps):
-            self.simulation.minimizeEnergy(tolerance=10**logc*kilojoule/mole)
+            self.simulation.minimizeEnergy(tolerance=10**logc*kilojoule/mole, maxIterations=10000)
+        # Remove the restraint energy from the total energy if desired.
+        groups = set(range(32))
+        if hasattr(self, 'restrain_k') and not include_restraint_energy:
+            for frc in self.simulation.system.getForces():
+                if isinstance(frc, CustomExternalForce):
```
yudongqiu: Getting the `frc` object can be improved by the same `frc = self.simulation.system.getForce(self.restraint_frc_index)` method described above.

src/openmmio.py
```diff
@@ -1117,34 +1168,68 @@ def normal_modes(self, shot=0, optimize=True):
...
+                    groups.remove(frc.getForceGroup())
+        # printcool_dictionary(energy_components(self.simulation), title='Energy component analysis after minimization, shot %i' % shot)
+        S = self.simulation.context.getState(getPositions=True, getEnergy=True, groups=groups)
         # Get the optimized geometry.
-        S = self.simulation.context.getState(getPositions=True, getEnergy=True)
         X1 = np.array([j for i, j in enumerate(S.getPositions().value_in_unit(angstrom)) if self.AtomMask[i]])
```
yudongqiu: I think this can be improved by
```
X1 = S.getPositions(asNumpy=True).value_in_unit(angstrom)[self. realAtomIdxs]
```

src/openmmio.py
```diff
@@ -1117,34 +1168,68 @@ def normal_modes(self, shot=0, optimize=True):
...
             M.align(center=False)
         X1 = M.xyzs[1]
         # Set geometry in OpenMM, requires some hoops.
         mod = Modeller(self.pdb.topology, [Vec3(i[0],i[1],i[2]) for i in X1]*angstrom)
-        mod.addExtraParticles(self.forcefield)
-        # self.simulation.context.setPositions(ResetVirtualSites(mod.getPositions(), self.system))
-        self.simulation.context.setPositions(ResetVirtualSites_fast(mod.getPositions(), self.vsinfo))
-        return E, M.ref_rmsd(0)[1]
+        if disable_vsite:
+            self.simulation.context.setPositions(mod.getPositions())
```
yudongqiu: I think this might be improved by
```
if disable_vsite:
    self.simulation.context.setPositions(X1 * angstrom)
else:
    mod = Modeller(self.pdb.topology, X1*angstrom)
    mod.addExtraParticles(self.forcefield)
    self.simulation.context.setPositions(ResetVirtualSites_fast(mod.getPositions(), self.vsinfo))
```
In this way we can skip creating the `mod` object when `disable_vsite==True`.

src/smirnoffio.py
```diff
@@ -629,6 +613,53 @@ def build_system_mval_masks(self):
...
+
+        Note
+        ----
+        1. This function assumes the names of the forcefield parameters has the smirks as the last item
+        2. This function assumes params only affect the smirks of its own. This might not be true if parameter_eval is used.
+        """
+        orig_pgrad_set = set(self.pgrad)
+        # smirks to param_idxs map
+        smirks_params_map = defaultdict(list)
+        for pname, pidx in self.FF.map.items():
```
yudongqiu: This is the old version to determine the mapping between smirks and affected param_idxs, I think we should use the updated version similar to other targets, like this:
```
        # New code for mapping smirks to mathematical parameter IDs
        for pname in self.FF.pTree:
            smirks = pname.rsplit('/',maxsplit=1)[-1]
            # print("pname %s mathid %s -> smirks %s" % (pname, str(self.FF.get_mathid(pname)), smirks))
            for pidx in self.FF.get_mathid(pname):
                smirks_params_map[smirks].append(pidx)
```

yudongqiu:
I haven't seen the implementation of `torsion_profile.py` yet, but I have a feeling that it reads the `metadata.json` file and freeze the atoms in the torsion been scanned.
One idea I have here is that is this worth to simplify the input of this target, by removing the requirement for the `metadata.json` file, and directly specify the torsion indices in the `$target` input block.
The idea is that all information we need for this target is the atom indices of the torsion. Since we will at most have 2 or 3 torsions being scanned at the same time, we may be able to directly load that information from the input file, so we don't need to worry about creating the `metadata.json` file in a specific structure.

src/torsion_profile.py
```diff
@@ -0,0 +1,191 @@
...
+class TorsionProfileTarget(Target):
+    """ Subclass of Target for fitting MM optimized geometries to QM optimized geometries. """
+    def __init__(self, options, tgt_opts, forcefield):
+        super(TorsionProfileTarget, self).__init__(options, tgt_opts, forcefield)
+
+        if not os.path.exists(os.path.join(self.tgtdir, 'metadata.json')):
+            raise RuntimeError('TorsionProfileTarget needs torsion drive metadata.json file')
+        with open(os.path.join(self.tgtdir, 'metadata.json')) as f:
+            self.metadata = json.load(f)
+            self.ndim = len(self.metadata['torsion_grid_ids'][0])
```
yudongqiu: I think it's more robust and easier to understand, if we load the dimension of torsion scan as `self.ndim = len(self.metadata['dihedrals'])`.

src/torsion_profile.py
```diff
@@ -0,0 +1,191 @@
...
+            self.freeze_atoms = sorted(list(set(itertools.chain(*self.metadata['dihedrals']))))
```
yudongqiu: A small improvement: `self.freeze_atoms = sorted(set(itertools.chain(*self.metadata['dihedrals'])))` works a bit faster.

src/torsion_profile.py
```diff
@@ -0,0 +1,191 @@
...
+        self.set_option(tgt_opts,'attenuate','attenuate')
+        ## Energy denominator for objective function
+        self.set_option(tgt_opts,'energy_denom','energy_denom')
+        ## Set upper cutoff energy
+        self.set_option(tgt_opts,'energy_upper','energy_upper')
+        ## Read in the reference data.
+        self.read_reference_data()
+        ## Build keyword dictionaries to pass to engine.
+        engine_args = OrderedDict(list(self.OptionDict.items()) + list(options.items()))
+        del engine_args['name']
```
yudongqiu: Small question: `del engine_args['name']` will raise `KeyError` if `engine_args` does not have `'name'` key. Is this what we want to check here? If not I think we can do
```
engine_args.pop('name', None)
```
So it won't raise Error.

src/torsion_profile.py
```diff
@@ -0,0 +1,191 @@
...
+                    self.wts[i] = 1.0 / np.sqrt(denom**2 + (eqm1[i]-denom)**2)
+        else:
+            self.wts = np.ones(self.ns)
+
+        # Normalize weights.
+        self.wts /= sum(self.wts)
+
+    def indicate(self):
+        title_str = "Torsion Profile: %s, Objective = % .5e, Units = kcal/mol, Angstrom" % (self.name, self.objective)
+        #QYD: This title is carefully placed to align correctly
```
yudongqiu: I guess this comment is copied from the `OptGeoTarget` 😀

src/torsion_profile.py
```diff
@@ -0,0 +1,191 @@
...
+        for line in open(os.path.join(self.root,self.qfnm)):
+            sline = line.split()
+            if len(sline) == 0: continue
+            elif sline[0] == 'ENERGY':
+                self.eqm.append(float(sline[1]))
+
+        if len(self.eqm) != self.ns:
+            raise RuntimeError("Length of qdata.txt should match number of structures")
+
+        # Turn everything into arrays, convert to kcal/mol
```
yudongqiu: I'm wondering if this is a good time to standardize the energy unit. Do we want to use `kcal/mol` everywhere, or is `kJ/mol` better?

src/torsion_profile.py
```diff
@@ -0,0 +1,191 @@
...
+        title_str = "Torsion Profile: %s, Objective = % .5e, Units = kcal/mol, Angstrom" % (self.name, self.objective)
+        #QYD: This title is carefully placed to align correctly
+        column_head_str1 =  "%-50s %-10s %-12s %-18s %-12s %-10s %-11s %-10s" % ("System", "Min(QM)", "Min(MM)", "Range(QM)", "Range(MM)", "Max-RMSD", "Ene-RMSE", "Obj-Fn")
+        printcool_dictionary(self.PrintDict,title=title_str + '\n' + column_head_str1, keywidth=50, center=[True,False])
+
+    def get(self, mvals, AGrad=False, AHess=False):
+        Answer = {'X':0.0, 'G':np.zeros(self.FF.np), 'H':np.zeros((self.FF.np, self.FF.np))}
+        self.PrintDict = OrderedDict()
+        # enable self.system_mval_masks (still need to implement)
+        enable_system_mval_mask = hasattr(self, 'system_mval_masks')
```
yudongqiu: These two lines are not needed here.

leeping:
This pull request **introduces 8 alerts** when merging 76fa39be630fc856ce1293502674cdb0a52daade into 27e7b4e433dd763ee1662d2ec409fa842caed545 - [view on LGTM.com](https://lgtm.com/projects/g/leeping/forcebalance/rev/pr-9bd12478510d0fd5fffae32a7e0dab8df4290921)

**new alerts:**

* 7 for Unused import
* 1 for Unused local variable

---

---

src/torsion_profile.py
```diff
@@ -0,0 +1,191 @@
...
+
+        def compute(mvals_, p_idx = None, indicate=False):
```
yudongqiu: the argument `p_idx` can be removed here

src/torsion_profile.py
```diff
@@ -0,0 +1,191 @@
...
+
+        self.PrintDict[self.name] = '%10s %10s    %6.3f - %-6.3f   % 6.3f - %-6.3f    %6.3f    %7.4f   % 7.4f' % (','.join(['%i' % i for i in self.metadata['torsion_grid_ids'][self.smin]]),
+                                                                                                                  ','.join(['%i' % i for i in self.metadata['torsion_grid_ids'][np.argmin(compute.emm)]]),
+                                                                                                                  min(self.eqm), max(self.eqm), min(compute.emm), max(compute.emm), max(compute.rmsd), e_rmse, Answer['X'])
+
+        # compute gradients and hessian
+        dV = np.zeros((self.FF.np,len(V)))
+        if AGrad or AHess:
+            for p in self.pgrad:
+                dV[p,:], _ = f12d3p(fdwrap(compute, mvals, p, p_idx = p), h = self.h, f0 = V)
```
yudongqiu: The argument `p_idx = p` can be removed here.

src/openmmio.py
```diff
@@ -1117,34 +1168,68 @@ def normal_modes(self, shot=0, optimize=True):
...
+        E : float
+            Potential energy of the system
+        rmsd : float
+            RMSD of the system (w/r.t. starting geometry) in Angstrom
+        """
 
         steps = int(max(1, -1*np.log10(crit)))
         self.update_simulation()
         self.set_positions(shot)
+        if hasattr(self, 'restrain_k'):
```
yudongqiu: I guess it's possible that `self.restrain_k` will get a default value of `0.0`. In this case we should skip this so maybe it's better to write
```
if hasattr(self, 'restrain_k') and self.restrain_k != 0:
```

src/openmmio.py
```diff
@@ -1117,34 +1168,68 @@ def normal_modes(self, shot=0, optimize=True):
...
         # Get the previous geometry.
         X0 = np.array([j for i, j in enumerate(self.simulation.context.getState(getPositions=True).getPositions().value_in_unit(angstrom)) if self.AtomMask[i]])
+        # printcool_dictionary(energy_components(self.simulation), title='Energy component analysis before minimization, shot %i' % shot)
         # Minimize the energy.  Optimizer works best in "steps".
         for logc in np.linspace(0, np.log10(crit), steps):
-            self.simulation.minimizeEnergy(tolerance=10**logc*kilojoule/mole)
+            self.simulation.minimizeEnergy(tolerance=10**logc*kilojoule/mole, maxIterations=10000)
+        # Remove the restraint energy from the total energy if desired.
+        groups = set(range(32))
+        if hasattr(self, 'restrain_k') and not include_restraint_energy:
```
yudongqiu: Same as above, we may want to check `if hasattr(self, 'restrain_k') and self.restrain_k != 0`

src/openmmio.py
```diff
@@ -1117,34 +1168,68 @@ def normal_modes(self, shot=0, optimize=True):
...
+            for frc in self.simulation.system.getForces():
+                if isinstance(frc, CustomExternalForce):
+                    groups.remove(frc.getForceGroup())
```
yudongqiu: I'm not exactly sure about this, but do we need to call `frc.setForceGroup()` when creating this force? I'm worried that if not, all forces will be in the same force group.

src/torsion_profile.py
leeping: Right, this problem is because the engine and target options both have the "name" key, and the target name should override the engine name.  A better design would have been to rename the Engine "name" key to something like "engine_name".  We should do that at some point.

src/torsion_profile.py
leeping: The rest of the comment is not wrong, so I should just change the initials. :)

src/torsion_profile.py
leeping: kJ/mol is kind of a legacy of abinitio.py which is a very old target, the first one.  I've since found that people tend to assume energies are in kcal/mol which is why I am going with it for torsion profiles.  We could start a new Github issue to discuss this, but I think that would quickly become a full blown thing folks will ask us to start requiring simtk.unit everywhere.  It could be worth considering but it's a lot of work.

yudongqiu: Yeah I agree sticking with one unit would good. If `kcal/mol` is the choice we can later make other target also use it. 👍

src/openmmio.py
leeping: Good point. I will make the replacement everywhere I check for this attribute.

src/openmmio.py
leeping: We already set the force group around line 840.  The only issue I can foresee is when something else overwrites the force group, for example when using a multiple timestep integrator in MD (around line 420).  However, this restraint force is not designed to be used with any target that runs MD at the moment.

yudongqiu: Oh sure I must have overlooked that part since it's not changed. No problem now.

src/openmmio.py
leeping: Changes have been made in the latest commit.

src/openmmio.py
leeping: Changes have been made in the latest commit.

src/openmmio.py
leeping: Changes have been made in the latest commit.

src/openmmio.py
leeping: Changes have been made in the latest commit.

src/openmmio.py
leeping: Changes have been made in the latest commit.

src/smirnoffio.py
leeping: Changes have been made in the latest commit.

src/torsion_profile.py
leeping: Changes have been made in the latest commit.

src/torsion_profile.py
leeping: Changes have been made in the latest commit.

src/torsion_profile.py
leeping: Changes have been made in the latest commit.

src/torsion_profile.py
leeping: Changes have been made in the latest commit.

src/openmmio.py
yudongqiu: Thanks! I forgot to mention that there is a same line above that can be improved in the same  way:
```
X0 = S.getPositions(asNumpy=True).value_in_unit(angstrom)[self.realAtomIdxs]
```

src/openmmio.py
leeping: I thought having this line everywhere was cumbersome so instead I check if self.restraint_frc_index is not None.

src/torsion_profile.py
leeping: Done.

leeping:
This pull request **introduces 2 alerts** when merging c8664cb3f361d07baf7ce8ae52b141d2c8f08b05 into 27e7b4e433dd763ee1662d2ec409fa842caed545 - [view on LGTM.com](https://lgtm.com/projects/g/leeping/forcebalance/rev/pr-12bfecd65596be2965c2c9a36e5ec9a720b7b52d)

**new alerts:**

* 1 for Unused local variable
* 1 for Unused import

---

---

src/openmmio.py
```diff
@@ -1168,10 +1247,7 @@ def multipole_moments(self, shot=0, optimize=True, polarizability=False):
             logger.error("Polarizability calculation is available in TINKER only.\n")
             raise NotImplementedError
 
-        if self.platname in ['CUDA', 'OpenCL'] and self.precision in ['single', 'mixed']:
-            crit = 1e-4
-        else:
-            crit = 1e-6
+        crit = 1e-4 if (self.platname in ['CUDA', 'OpenCL'] and self.precision in ['single', 'mixed']) else 1e-6
```
j-wags: Stylistically, this was a bit easier to understand before the change.

src/openmmio.py
```diff
@@ -1191,14 +1267,11 @@ def energy_rmsd(self, shot=0, optimize=True):
 
         self.update_simulation()
 
-        if self.platname in ['CUDA', 'OpenCL'] and self.precision in ['single', 'mixed']:
-            crit = 1e-4
-        else:
-            crit = 1e-6
+        crit = 1e-4 if (self.platname in ['CUDA', 'OpenCL'] and self.precision in ['single', 'mixed']) else 1e-6
```
j-wags: Same style comment as above

src/smirnoffio.py
```diff
@@ -667,6 +653,55 @@ def build_system_mval_masks(self):
         logger.info("system_mval_masks is built for faster gradient evaluations")
         self.system_mval_masks = system_mval_masks
 
+class TorsionProfileTarget_SMIRNOFF(TorsionProfileTarget):
+    """ Force and energy matching using SMIRKS native Open Force Field (SMIRNOFF). """
+    def __init__(self,options,tgt_opts,forcefield):
+        ## Default file names for coordinates and key file.
+        self.set_option(tgt_opts,'pdb',default="conf.pdb")
```
j-wags: If this PDB is only being created for the purpose of loading coordinates using the OFF toolkit, it may be possible to replace it with coordinates loaded from mol2 (as long as only a single molecule is being loaded at a time)

src/smirnoffio.py
```diff
@@ -667,6 +653,55 @@ def build_system_mval_masks(self):
...
+        # List of .mol2 files for SMIRNOFF to set up the system
+        self.set_option(tgt_opts,'mol2',forceprint=True)
```
j-wags: The OFF toolkit currently only "trusts" OpenEye to load mol2 files, so if this is calling the OFF toolkit, this makes OpenEye a dependency.

src/openmmio.py
leeping: I've reverted this in the latest commit.

src/openmmio.py
leeping: I've reverted this in the latest commit.

src/smirnoffio.py
leeping: The PDB file is also used for adding fields to the `forcebalance.molecule` object which is used through FB in several places, so I think it's best to keep it.

src/smirnoffio.py
leeping: Noted. If we want to use RDKit, should we load templates for individual molecules using .sdf?

leeping:
Thank you for the comments!  I'm merging the PR.

leeping:
This pull request **introduces 2 alerts** when merging 61e815070d7aaebac0bede70e1ff9047eb0fcb22 into 27e7b4e433dd763ee1662d2ec409fa842caed545 - [view on LGTM.com](https://lgtm.com/projects/g/leeping/forcebalance/rev/pr-f1a02f50365465390baac7b4c57f62474aeb2e61)

**new alerts:**

* 1 for Unused local variable
* 1 for Unused import

---

---

src/smirnoffio.py
j-wags: That's right. RDKit has some trouble with loading charges from SDF, but I can send you a code snippet to handle that until we work out a better solution.

PR #156, opened 2019-08-13
Recent Improvements related to SMIRNOFF fitting

yudongqiu:
This PR contains several improvements related to SMIRNOFF fitting:

1. Add writing the `"EnergyCompare.txt"` file in the `TorsionProfile` target. This file adopts the same format as the `AbInitio` target and is very useful for analyzing the fitting results.

2. Improve the stability of the OpenMM engine `optimize()` function. This improvement solves two problems. a) When the `minimizeEnergy()` does not converge within the `maxIterations`, it will raise error and let the user know. b) Even if the `simulation.minimizeEnergy()` converged, I found that the energy could still go down when re-run the function. Now we try that 10 times and check if the energy "really converged" or not.

3. The `smirnoff_analyze_parameter_coverage()` function is re-written. Now it takes the target options dict, instead of the created Target objects. A benefit is we don't need to worry about the target being "remote" or not. The file `smirnoff_parameter_assignments.json` created is also improved, to use path of each `mol2` file as a key instead of the target name. In this way, we can read the json file for the mapping of SMIRKs terms for each molecule.

4. The `allow_undefined_stereo=True` argument is removed from all functions like `OffMolecule.from_file()` which uses the `openforcefield` toolkit to load mol2 file. This is suggested by @j-wags as it's not completely safe to use molecules with "undefined stereo" in fitting.

### DIFF
--- a/src/openmmio.py
+++ b/src/openmmio.py
@@ -619,7 +619,7 @@ class OpenMM(Engine):
 
         # Store a separate copy of the molecule for reference restraint positions.
         self.ref_mol = deepcopy(self.mol)
-        
+
     def prepare(self, pbc=False, mmopts={}, **kwargs):
 
         """
@@ -733,7 +733,7 @@ class OpenMM(Engine):
 
         ## Generate list of OpenMM-compatible positions
         self.generate_xyz_omm(self.mol)
-            
+
         ## Build a topology and atom lists.
         Top = mod.getTopology()
         Atoms = list(Top.atoms())
@@ -830,10 +830,10 @@ class OpenMM(Engine):
 
         # Freeze atoms if we have any.
         if hasattr(self, 'freeze_atoms'):
-            for i, j in enumerate(self.realAtomIdxs):
-                if i in self.freeze_atoms:
-                    self.system.setParticleMass(j, 0.0)
-                        
+            for i in self.freeze_atoms:
+                j = self.realAtomIdxs[i]
+                self.system.setParticleMass(j, 0.0)
+
         ## Set up for energy component analysis.
         GrpTogether = ['AmoebaGeneralizedKirkwoodForce', 'AmoebaMultipoleForce','AmoebaWcaDispersionForce',
                         'CustomNonbondedForce',  'NonbondedForce']
@@ -904,7 +904,7 @@ class OpenMM(Engine):
             if isinstance(i, AmoebaMultipoleForce):
                 if self.SetPME:
                     i.setNonbondedMethod(i.PME)
-                    
+
         #----
         # If the virtual site parameters have changed,
         # the simulation object must be remade.
@@ -1158,10 +1158,10 @@ class OpenMM(Engine):
 
     def optimize(self, shot, crit=1e-4, disable_vsite=False, align=True, include_restraint_energy=False):
 
-        """ 
-        Optimize the geometry and align the optimized 
+        """
+        Optimize the geometry and align the optimized
         geometry to the starting geometry.
-        
+
         Parameters
         ----------
         shot : int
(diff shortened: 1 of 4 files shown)


src/smirnoffio.py
```diff
@@ -58,49 +58,38 @@
...
+            optgeo_options_txt = os.path.join(target_path, tgt_option['optgeo_options_txt'])
+            sys_opts = forcebalance.opt_geo_target.OptGeoTarget.parse_optgeo_options(optgeo_options_txt)
+            mol2_paths = [os.path.join(target_path,fnm) for sysopt in sys_opts.values() for fnm in sysopt['mol2']]
+        elif tgt_option['type'].endswith('_SMIRNOFF'):
+            mol2_paths = [os.path.join(target_path,fnm) for fnm in tgt_option['mol2']]
+        # analyze SMIRKs terms
+        for mol_fnm in mol2_paths:
+            # we work with one file at a time to avoid the topology sliently combine "same" molecules
+            openff_mols = [OffMolecule.from_file(mol_fnm)]
+            off_topology = OffTopology.from_molecules(openff_mols)
```
j-wags: As a small detail, I'd move defining the list `[openff_mols]` to inside of `OffTopology.from_molecule([openff_mol])`, to make it more consistent with how it's used in other consortium code. This is a style suggestion and you can feel free to ignore it.

src/smirnoffio.py
```diff
@@ -58,49 +58,38 @@
...
+            # parse optgeo_options_txt and get the names of the mol2 files
+            optgeo_options_txt = os.path.join(target_path, tgt_option['optgeo_options_txt'])
+            sys_opts = forcebalance.opt_geo_target.OptGeoTarget.parse_optgeo_options(optgeo_options_txt)
+            mol2_paths = [os.path.join(target_path,fnm) for sysopt in sys_opts.values() for fnm in sysopt['mol2']]
+        elif tgt_option['type'].endswith('_SMIRNOFF'):
+            mol2_paths = [os.path.join(target_path,fnm) for fnm in tgt_option['mol2']]
+        # analyze SMIRKs terms
+        for mol_fnm in mol2_paths:
+            # we work with one file at a time to avoid the topology sliently combine "same" molecules
+            openff_mols = [OffMolecule.from_file(mol_fnm)]
```
j-wags: If you want to implement non-fatal logic when a bad molecule is loaded, you could include a try/except like 
```
try:
    openff_mols = [OffMolecule.from_file(mol_fnm)]
except openforcefield.utils.toolkits.UndefinedStereochemistryError as e:
    logger.warning(e.msg)
```

src/smirnoffio.py
```diff
@@ -283,7 +272,7 @@ def prepare(self, pbc=False, mmopts={}, **kwargs):
         openff_mols = []
         for fnm in self.mol2:
             try:
-                mol = OffMolecule.from_file(fnm, allow_undefined_stereo=True)
+                mol = OffMolecule.from_file(fnm)
             except Exception as e:
```
j-wags: See earlier comment for more specific `UndefinedStereochemistryError` class path

src/openmmio.py
```diff
@@ -1191,7 +1191,18 @@ def optimize(self, shot, crit=1e-4, disable_vsite=False, align=True, include_res
         # printcool_dictionary(energy_components(self.simulation), title='Energy component analysis before minimization, shot %i' % shot)
         # Minimize the energy.  Optimizer works best in "steps".
         for logc in np.linspace(0, np.log10(crit), steps):
-            self.simulation.minimizeEnergy(tolerance=10**logc*kilojoule/mole, maxIterations=10000)
+            self.simulation.minimizeEnergy(tolerance=10**logc*kilojoule/mole, maxIterations=100000)
+        # check if energy minimization is successful
+        # try 10 times as openmm minimizer is not very stable at the tolerance
```
j-wags: Oh wow. That's very unfortunate. I'm not very experienced with OpenMM, but I did find cases where the `Reference` platform gave me inconsistent results, while the `OpenCL` platform was fully consistent between reruns of the same system. If these retries add a significant time cost increase, hard-coding the platforms may help us in the run-up to release 1. It would, however, make the code less general.

src/smirnoffio.py
yudongqiu: Sure it's updated. The `openff_mols` list was used to aggregate all related `mol2` files into a `off_topology`, and run `label_molecules(off_topology)`. I later realized that the `off_topology` automatically removes the duplicated molecules, which I don't want. Therefore I changed that to work with one `mol2` file at a time.

src/smirnoffio.py
yudongqiu: ForceBalance assumes all targets are "good", and the quality control for targets is done before we run ForceBalance, so I think we would want to let it raise an error here.

src/openmmio.py
yudongqiu: I agree it's worth revisiting the problem later and probably raise an issue for OpenMM. I think trying to re-run the minimizeEnergy here is working for now.

src/smirnoffio.py
j-wags: Sounds good.

src/smirnoffio.py
j-wags: Great. Thanks for the explanation -- that makes sense.

PR #162, opened 2019-10-16
save mm optimized geometry from optgeo targets

hyejang:
Small change has been made in opt_geo_target.py to save the MM minimized geometries for optgeo targets when writelevel is greater than zero. This writing will not happen when evaluating numerical gradients.


leeping:
@hyejang Could you please merge with master and push to this branch again?

leeping (approved):

leeping:
Thanks a lot - merging (finally).

PR #163, opened 2019-10-23
wip Convert to pytest format

hyejang:


src/tests/test_continue.py
```diff
@@ -12,62 +12,97 @@
...
-        super(ForceBalanceTestCase,self).setUp()
-        os.system('rm -rf test/files/test_continue.run')
-        os.makedirs('test/files/test_continue.run')
-        os.system('cp -r studies/001_water_tutorial/forcefield test/files/test_continue.run/forcefield')
-        os.system('cp -r studies/001_water_tutorial/targets test/files/test_continue.run/targets')
-        os.chdir('test/files')
-        os.system('cp -r test_continue.sav test_continue.in test_continue.tmp test_continue.run')
-        os.chdir('test_continue.run')
+# class TestWaterTutorial(ForceBalanceTestCase):
+class TestWaterTutirual:
```
j-wags:
```suggestion
class TestWaterTutorial:
```

src/tests/test_forcefield.py
```diff
@@ -16,33 +14,30 @@ class FFTests(object):
...
-        msg = "\nGot two different forcefields despite using the same options as input")
+        print("\nChecking consistency of ForceField constructor\n")
+        assert forcefield.FF(self.options) == forcefield.FF(self.options), "Got two different forcefields despite using the same options as input"
 
     def test_make_function_return_value(self):
         """Check that make() return value meets expectation"""
         pvals = self.ff.pvals0
 
-        self.logger.debug("Running forcefield.make() with zero vector should not change pvals... ")
+        print("Running forcefield.make() with zero vector should not change pvals... ")
```
j-wags: Can we continue to use a logger? Where would it be defined?

src/tests/test_finite_difference.py
```diff
@@ -1,64 +1,60 @@
...
+#     @classmethod
+# def setup_class(cls):
+    # functions is a list of 3-tuples containing (function p, d/dp, d^2/dp^2)
+functions = []
 
-class TestFiniteDifference(ForceBalanceTestCase):
-    def setUp(self):
-        # functions is a list of 3-tuples containing (function p, d/dp, d^2/dp^2)
-        self.functions = []
+# f(x) = 2x
```
j-wags: Can we continue to have things organized in classes? Do we lose anything essential by removing ForceBalanceTestCase? Can we replace it with equivalent functionality from pytest?

src/tests/test_finite_difference.py
```diff
@@ -1,64 +1,60 @@
...
+# class TestFiniteDifference(ForceBalanceTestCase):
+# class TestFiniteDifference:
+#     # def setUp(self):
+#     @classmethod
+# def setup_class(cls):
+    # functions is a list of 3-tuples containing (function p, d/dp, d^2/dp^2)
+functions = []
 
-class TestFiniteDifference(ForceBalanceTestCase):
-    def setUp(self):
```
j-wags: Do we gain anything by having `setUp` run in a class, rather than defining important variables for our tests at the module level? Is there anything bad about defining these at the module level?

src/tests/test_continue.py
```diff
@@ -12,62 +12,97 @@
...
+    # def setUp(self):
+    @classmethod
+    def setup_class(cls):
```
j-wags: Do we want to use `setup_class`, or a different setup that will run for each test (rather than all the tests in the same class)?

src/tests/test_continue.py
```diff
@@ -12,62 +12,97 @@
...
+        # os.system('cp -r test_continue.sav test_continue.in test_continue.tmp test_continue.run')
+        shutil.copy("test_continue.sav", tmpfolder)
+        shutil.copy("test_continue.in", tmpfolder)
+        shutil.copytree("test_continue.tmp", os.path.join(tmpfolder,"test_continue.tmp"))
+        os.chdir(tmpfolder)
+
+    # def tearDown(self):
+    @classmethod
+    def teardown_class(cls):
+        # os.chdir('..')
```
j-wags: Is it OK that we don't change directories during teardown? Do the tests end up in the wrong place if we don't return to the same directory after each one?

src/tests/test_continue.py
j-wags: I think it's best to use `setup_method`, so that setup and teardown happens _between each test_ rather than _before or after all of the tests_

src/tests/test_continue.py
j-wags: We should change the `setup` function that runs before tests to remember which directory it starts in (`self.start_directory = os.getcwd()`) and then return to that directory during teardown (`os.chdir(self.start_directory)`). Otherwise it will be easy to get lost in directories if we always _assume_ that we go down two directories before a test, and then back up two directories after it. By defining an absolute path, we avoid potentially making that mistake.

src/tests/test_forcefield.py
```diff
@@ -16,33 +14,30 @@ class FFTests(object):
...
-        self.logger.debug("ok\n")
+        assert not (pvals==new_pvals).all(), "make() returned unchanged pvals even when given nonzero matrix"
+        print("ok\n")
 
-        self.logger.debug("make(use_pvals=True) should return the same pvals... ")
+        print("make(use_pvals=True) should return the same pvals... ")
         new_pvals = np.array(self.ff.make(np.ones(self.ff.np),use_pvals=True))
-        self.assertEqual(np.ones(self.ff.np),new_pvals, msg="\nmake() did not return input pvals with use_pvals=True")
-        self.logger.debug("ok\n")
+        assert (np.ones(self.ff.np) == new_pvals).all(), "make() did not return input pvals with use_pvals=True"
```
j-wags: `==` may not be an appropriate replacement for all uses of `self.assertEqual`, since `unittest.assertEqual` has its behavior overridden specifically for numpy arrays in `__init__`. So we should use something like Hyesu's 
`np.testing.assert_array_almost_equal(EXPECTED_WATER_RESULTS,result,decimal=0.001, err_msg=msg) `
below to explicitly describe how we're doing `np.array` comparisons (and with what tolerance)

src/tests/test_finite_difference.py
j-wags: We can keep things organized in classes. See the changes in `test_tinkerio` and `__init__`.

src/tests/test_finite_difference.py
j-wags: We should use `setup_method` and `teardown_method` wherever possible.

src/tests/test_forcefield.py
j-wags: The logger is now part of the `ForceBalanceTestCase` class definition (defined in `setup_class`). So we can revert things to `self.logger`

src/tests/test_continue.py
leeping: There are cases where it makes sense to have setup and teardown such that the workflow goes as setup-test1-test2-teardown, as opposed to setup1-test1-teardown1-setup2-test2-teardown2.

For example, if we set up a system in GROMACS and TINKER, and see if a number of properties match (e.g. gradients, optimized geometry, and vibrational frequency).  Because each property involves running different parts of the codes, it makes sense to have them as different tests.  However, it takes a small amount of time to set up the system so it makes sense to setup and teardown only once.

A related issue is that for the tests above, they should not have separate setup and teardown codes because that would introduce redundancy.  This may be resolvable separately from the computational cost concern.

src/tests/test_continue.py
leeping: Yes, I agree with you.  The FB process does change folders as it runs, and that was a design choice I made early on (for better or worse) because it has to create and manage a complex directory structure.  We should have a unified condition for all tests that they must finish in the same folder that they're started in.

src/tests/test_finite_difference.py
leeping: The possible drawback of defining setup and teardown at the module level is if we want to run the following sequence of tests in a single module (which happens most often in test_engine.py):

setupA-testA1-testA2-teardownA (e.g. test optimized geometries and vibrational frequencies for Gromacs vs. AMBER for the alanine-glutamate system)
setupB-testB1-testB2-teardownB (e.g. test optimized geometries and vibrational frequencies for Gromacs vs. AMBER for the water cluster system)

src/tests/test_finite_difference.py
leeping: I see this comment was made after the others, so I'm interested in knowing how you keep things organized in classes.  I do see value in it (from comments above) but it's worth discussing whether the value is worth the extra "bureaucracy".

src/tests/test_forcefield.py
leeping: Thanks. There is a logger in the main ForceBalance codes and I'd like to know how the test logger differs / is similar to the main one. On a side note I need a refresher on why `src/logger.py` was designed that way, and write some good docs to explain them.

src/tests/test_forcefield.py
leeping: Yes, I think that maybe it's best to choose the proper comparison function or operator on a case-by-case basis rather than defining a catch-all comparison function and trying to use it everywhere.

src/nifty.py
```diff
@@ -1406,10 +1406,10 @@ def process_err(read):
                     read = ''.encode('utf-8')
                     while True:
                         if read_nbytes == 0:
-                            read += fh.read(rbytes)
+                            read += fh.read1(rbytes)
                             read_nbytes += rbytes
                         else:
-                            read += fh.read(1)
+                            read += fh.read1(1)
```
leeping: Very interesting. :) What does this do?

src/tests/test_finite_difference.py
j-wags: I'm pretty new to test structure as well. Generally I just use classes for bookkeeping -- The use of inheritance from `ForceBalanceTestCase` is interesting, and is probably worth keeping around. I'm less certain about `ForceBalanceTestResult` and `ForceBalanceTestRunner` -- I haven't read too closely, but the functionality there may be redundant with what's built in to pytest.

The "bureaucracy" goes both ways. There have been times where the act of grouping tests has made me find redundancies or gaps in coverage around certain functionalities. The FB tests seem very well thought-out, so there are probably fewer of that type of inconsistency. I could go either way on the use of classes, and removing them is as simple as changing `setup_class` to `setup_module` and `setup_method` to `setup_function`.

src/nifty.py
j-wags: It solves the OSX hanging problem by having it only attempt to read the stream once. My guess is that, if `read` is used instead of `read1`, it will repeatedly attempt the read until the desired number of bytes (default here is 1024) come out. So, there's either a difference between the `read` function on mac vs linux, or `Tinker` may have slightly different behavior.

Either way, reducing `rbytes` from `1024` to some value closer to `500` fixes it on mac (probably because the total output size is less than 1024 bytes in some cases), but I wasn't very satisfied with that solution. After some more digging, I found [`read1`](https://docs.python.org/3/library/io.html#io.BufferedIOBase.read1), which attempts to get the desired number of bytes using a single read operation, and just returns what it was able to get otherwise. 

I'm interested in getting the test matrix running again so we can assess if this sneakily broke something else in the process.

.travis.yml
```diff
@@ -3,37 +3,65 @@
...
-python:
-    - 2.7
-    - 3.5
-    - 3.6
+matrix:
+  include:
+  # Test osx
+  - os: osx
+    language: generic
+    env: PYTHON_VER=2.7
```
j-wags: This will get into a lock state due to the lack of `ByteStream.read1` in py27. Remove py27 from mac tests

src/tests/test_continue.py
```diff
@@ -13,61 +12,71 @@
...
+        super().teardown_method()
+
+    def test_continue(self):
         """Check continuation from a previous run"""
-        if not sys.version_info <= (2,7):
-            self.skipTest("Existing pickle file only works with Python 3")
-        self.logger.debug("\nSetting input file to 'test_continue.in'\n")
+        # if not sys.version_info <= (2,7):
+        #     skipTest("Existing pickle file only works with Python 3")
+        self.logger.debug("Setting input file to 'test_continue.in'")
```
j-wags: `logger.debug` doesn't print newlines automatically -- Add them here

src/tests/test_continue.py
```diff
@@ -13,61 +12,71 @@
...
+        if os.path.isdir(tmpfolder):
+            shutil.rmtree(tmpfolder)
+        super().teardown_method()
+
+    def test_continue(self):
         """Check continuation from a previous run"""
-        if not sys.version_info <= (2,7):
-            self.skipTest("Existing pickle file only works with Python 3")
-        self.logger.debug("\nSetting input file to 'test_continue.in'\n")
+        # if not sys.version_info <= (2,7):
```
j-wags: Note to self -- Could this help skip inappropriate tests for mac py27?

src/tests/test_engine.py
```diff
@@ -59,8 +60,8 @@ class TestAmber99SB:
     Vibrational frequencies: < 0.5 wavenumber (~ 1e-4 fractional error)
     Vibrational eigenvectors: < 0.01
     """
-    @classmethod
-    def setup_class(cls):
+    def setup_method(self, method):
+        super().setup_method(method)
         """
         setup any state specific to the execution of the given class (which usually contains tests).
         """
```
j-wags: Let's change this back to setup_class, and move the super() below the docstring

src/tests/test_engine.py
```diff
@@ -70,40 +71,41 @@ def setup_class(cls):
...
-    def teardown_class(cls):
+    def teardown_method(self):
+
         """
         teardown any state that was previously setup with a call to setup_class.
         """
-        os.chdir('..')
-        shutil.rmtree('temp')
+        os.chdir(self.cwd)
+        shutil.rmtree(self.cwd, "files", "amber_alaglu", "temp")
```
j-wags: For now, let's keep test output (so don't remove files). If this causes subsequent tests to fail, we can reintroduce it. And later we can see about implementing cleanup conditional on tests passing.

src/tests/test_engine.py
```diff
@@ -327,32 +329,33 @@ class TestAmoebaWater6:
...
+        self.O = OpenMM(coords="hex.arc", pdb="prism.pdb", ffxml="amoebawater.xml", precision="double", \
                             mmopts={'rigidWater':False, 'mutualInducedTargetEpsilon':1e-6})
         tinkerpath = which('testgrad')
         if tinkerpath:
-            cls.T = TINKER(coords="hex.arc", tinker_key="prism.key", tinkerpath=tinkerpath)
+            self.T = TINKER(coords="hex.arc", tinker_key="prism.key", tinkerpath=tinkerpath)
 
     @classmethod
-    def teardown_class(cls):
+    def teardown_class(self):
```
j-wags: Let's take a close look at these and determine which lines should run in test_class and which should run in test_method.

src/tests/test_molecule.py
```diff
@@ -76,13 +81,15 @@ def test_gro_conversion(self):
             self.fail("\nConversion to gro format creates unreadable file")
 
         self.logger.debug("\nChecking that conversion has not changed number of residues\n")
-        self.assertEqual(len(self.molecule.Data['resid']), len(molecule1.Data['resid']),
-                        msg = "\nConversion from pdb to gro yields different number of residues")
+        assert len(self.molecule.Data['resid']) == len(molecule1.Data['resid'])
```
j-wags: Jeff needs to put the assert message here

devtools/conda-envs/test_env.yaml
```diff
@@ -0,0 +1,27 @@
...
+channels:
+  - conda-forge
+  - omnia
+  - ambermd
+dependencies:
+    # Base depends
+  - python
+  - pip
+
+    # Testing
```
j-wags: Add openforcefield

j-wags: Add openeye-toolkits and openeye, configure OE license. List instructions for how to update when it expires.

PR #157, opened 2019-08-21
Property Estimator

yudongqiu:
This PR contains collaborative effort from @SimonBoothroyd and I. The interface between [PropertyEstimator](https://github.com/openforcefield/propertyestimator) was implemented in 

1. A mock target design: https://github.com/yudongqiu/forcebalance/pull/5
2. The full implementation: https://github.com/yudongqiu/forcebalance/pull/7

PR #164, opened 2019-11-11
Update PropertyEstimator API Calls

SimonBoothroyd:
## Description
This PR is an extension to #157, and aims to:

- change use to the [new API syntax](https://property-estimator.readthedocs.io/en/0.0.5/releasehistory.html#breaking-changes) introduced in propertyestimator version 0.0.4
- add the `024_property_estimator` studies `targets` directory which was previously missing 

## TODO

- [x] Re-run study  `024_property_estimator` to ensure that the target now works correctly.

## Status
- [x] Ready to go


SimonBoothroyd:
Currently this [PR](https://github.com/leeping/forcebalance/blob/2c5891d37277ffecb0ff747ee2ed67b4994b78de/src/property_estimator_io.py#L336) and [ForceBalance](https://github.com/leeping/forcebalance/blob/891bab94d4328fb5a011d532f643ba306690419d/src/smirnoffio.py#L156) relies on accessing an internal field of the toolkit. We should make sure to change this when https://github.com/openforcefield/openforcefield/issues/393 is closed.

leeping:
I'll merge and rerun the tests on master so that it properly runs the propertyestimator test.  Thanks, Simon!

PR #165, opened 2019-11-24
Use gradient evaluation for Evaluator test

leeping:
- Pinned versions of pymbar and mdtraj (until propertyestimator aka Evaluator release pins these versions)
- Do not satisfy step-size convergence criterion if returning to a previous point after a bad step
- Refactor test_system.py, removing duplicate codes
- Add `gradient.in` for testing Evaluator gradients (results have a large uncertainty)
- Performance optimization of Evaluator test


leeping:
Sounds good, thanks for all of your help!

PR #171, opened 2020-02-11
Test OpenEye is Licensed

SimonBoothroyd:
## Description
This PR updates the travis configuration to check that `openeye` is both importable and licensed when it is being installed so that tests can exit early when this isn't the case.

(OE not being licensed appears to be the source of the current test failures on `master`)

## Status
- [x] Ready to go


leeping:
Thank you, this looks good to me.

j-wags:
Oh, this is really nifty. Thanks for tagging me, @SimonBoothroyd!

PR #173, opened 2020-03-03
[Evaluator] Fix Exception Raised for Undefined Uncertainties

SimonBoothroyd:
## Description

Fitting against properties which do not have uncertainties defined currently leads to an exception being raised.

This PR adds checks for these cases and simply sets the reported uncertainty to be nan when not present.

## Status
- [X] Ready to go

### DIFF
--- a/src/evaluator_io.py
+++ b/src/evaluator_io.py
@@ -20,6 +20,7 @@ from forcebalance.target import Target
 
 try:
     from evaluator import unit
+    from evaluator.attributes import UNDEFINED
     from evaluator.client import EvaluatorClient, ConnectionOptions, RequestOptions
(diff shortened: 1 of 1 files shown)


leeping:
It seems like Python 2.7 tests are failing.  I'm not sure if this is stochastic .. could you please check?

SimonBoothroyd:
> It seems like Python 2.7 tests are failing. I'm not sure if this is stochastic .. could you please check?

Hmm it seems like the tests in `master` are also failing, so I think failure is probably real, but unrelated to this PR. I'll have a bit of a closer look.

SimonBoothroyd:
I opened an issue for this failure here: #174

PR #182, opened 2020-06-23
Update for OFFTK 070 

j-wags:
Many cached methods in `smirnoff_hack` were broken by the 0.7.0 OFF Toolkit release. This PR fixes the existing OpenFF ToolkitWrapper caching methods, and adds caching for other expensive RDKit/AmberTools/OETK ToolkitWrapper methods.

@leeping, since these changes will be reverse-incompatible, we might choose to document the OFFTK>=0.7.0 version requirement somewhere. Is there anywhere in the docs/build instructions that would fit?


hyejang:
@j-wags, while looking into the changes you made in `smirnoff_hack.py`, I found that in line 59 and 62, there are print functions for debugging. We would want to remove the lines before merging into master.

j-wags:
My mistake. Fixed!

leeping:
Thank you, merging. :)

PR #187, opened 2020-08-04
Fix Parsing of Parameter Eval Statements with Commas

### LINKED ISSUE #186: Parameter Eval with Comma Containing SMIRKS Patterns
## Description

I am trying to use the `parameter_eval` flag with a SMIRNOFF force field to set the value of one parameter attribute equal to the negative value of a second:

```
<ChargeIncrement smirks="[#6X4:1]-[#8X1,#8X2:2]" charge_increment1="0.0718 * elementary_charge"
                         charge_increment2="-0.0718 * elementary_charge"
                         parameterize="charge_increment1"
                         parameter_eval="charge_increment2=-PRM['ChargeIncrementModel/ChargeIncrement/charge_increment1/[#6X4:1]-[#8X1,#8X2:2]']"/>
```

This works well if the SMIRKS pattern does not contain a comma (such as in the case of `[#6X4:1]-[#1:2]`) but fails for those SMIRKS patterns (such as `[#6X4:1]-[#8X1,#8X2:2]` in the above example) that do as this is also the delimiter ForceBalance uses to separate eval statements.

Likely a more sophisticated split pattern is required in `FF.addff_xml` to ensure that any text within the `PRM[...]` block are not split.

(Possibly a PR to follow)

SimonBoothroyd:
## Description

The PR switches to using a simple regex pattern which ignores text between [ ... ] brakets to split the list of `parameter_eval` statements. This allows eval statements which contain nested commas inside of the PRM quotes (e.g. which contain SMIRKS patterns which themselves contain commas) to be correctly split.

Closes #186

### DIFF
--- a/src/forcefield.py
+++ b/src/forcefield.py
@@ -97,6 +97,7 @@ from __future__ import division
 from builtins import zip
 from builtins import str
 from builtins import range
+import csv
 import os
 import sys
 import numpy as np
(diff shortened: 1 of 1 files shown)


SimonBoothroyd:
Unfortunately `csv` will not work here, as it seems to only work properly when the quoted text not to split is directly preceded and proceeded by a comma.

leeping:
Looks good to me, and merging!  Thanks.

PR #194, opened 2020-12-14
Update toolkit calls for new Molecule inheritance

### LINKED ISSUE #193: Not Compatable with Openforcefield toolkit 0.8.1
In the recent openforcefield toolkit release 0.8.1 some extra arguments were added to some of the [functions](https://github.com/openforcefield/openforcefield/blob/4641ec15bce760e6e3ad0e15ead70dcf3191c361/openforcefield/utils/toolkits.py#L4378) which are currently replaced in this [file](https://github.com/leeping/forcebalance/blob/master/src/smirnoff_hack.py) to enable using a cache.  I think these functions now need to be updated to allow the passing of the extra arguments and it may be possible to future proof this issue using *args and **kwargs. 

Error message
`ToolkitWrapper around AmberTools version 20.0 <class 'TypeError'> : at_cached_assign_partial_charges() got an unexpected keyword argument '_cls'
`

 This error is shown when using forcebalance1.7.5, openforcefield0.8.1, ambertools 20.9

mattwthompson:
Fixes #193 

In 0.8.1, `Molecule` methods that call out to toolkit wrapper functions (like `Molecule.assign_partial_charges`) and the toolkit functions themselves (like `AmberToolsToolkitWrapper.assign_partial_charges`) had their function signatures updated to require a `_cls` argument. This change to the toolkit was made to enable Molecule subclasses to function. In the scope of ForceBalance, I don't think that these changes

The gory details are here: https://github.com/openforcefield/openforcefield/pull/583, but to briefly summarize:
* The `_cls` argument defaults to `Molecule` in the Molecule API, but defaults to `None` in the toolkit APIs
  * There may have been a good reason to not default these to `Molecule` as well, but I don't remember off the top of my head.
* This change only affects toolkit wrapper methods that involve making a new `Molecule` instance, since somebody subclassing out of it may want those methods to act on their subclass, not molecule, so directly calling `Molecule()` in those functions was causing problems.
  * I don't think `.find_smarts_matches` involves making a new molecule, so those functions don't need to be changed.

I'm not sure how best to test this; I couldn't find this tested in the unit tests here. With this patch, I can locally type a topology with each toolkit, whereas I cannot on the most recent master branch. There may be more complexity I have not considered, but this gets the minimum use case working. This can be wrapped into a unit test easily enough if this is the right behavior to test.

```
from openforcefield.topology import Molecule, Topology
from openforcefield.typing.engines.smirnoff import ForceField
from openforcefield.utils.toolkits import (AmberToolsToolkitWrapper,
                                           OpenEyeToolkitWrapper,
                                           RDKitToolkitWrapper,
                                           ToolkitRegistry)

from forcebalance import smirnoff_hack

top = Topology.from_molecules(Molecule.from_smiles("CCO"))
parsley = ForceField("openff-1.0.0.offxml")

for toolkit_wrapper_class in [
    OpenEyeToolkitWrapper,
    RDKitToolkitWrapper,
    AmberToolsToolkitWrapper,
]:
    registry = ToolkitRegistry()
    registry.register_toolkit(toolkit_wrapper_class)
    if toolkit_wrapper_class == RDKitToolkitWrapper:
        registry.register_toolkit(AmberToolsToolkitWrapper)
    out = parsley.create_openmm_system(top, toolkit_registry=registry)


mattwthompson:
Alright, now that CI is back online, I've updated this to include a simple test. It's pretty much the snippet above, except without testing OpenEye (there should be no need) and skipping if `openforcefield` is not installed. This passed locally in an environment with it installed and skipped it in an environment with it not installed; hopefully it behaves the same in CI.

The migration associated with the toolkit's namespace change can take place in this PR if desired; however, I think it would be better to do that in a separate PR.

leeping (approved):

PR #195, opened 2020-12-14
Update tinker version in Travis CI

mattwthompson:
I'm not sure if any other changes happened with recent versions of Tinker; all this does is update the URL to point to files that still exist on the download page.

By the way, if travis continues to be slow to churn through jobs, it may be related to their recent efforts to move away from supporting open source software. It may be worth considering a move to GitHub Actions or Azure Pipelines, which have been more generous to open source projects in recent years.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -125,15 +125,15 @@ install:
   - which mdrun_d
   
     # Tinker binaries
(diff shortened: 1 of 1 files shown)


leeping:
Thank you, Matt.  Yes, I'd be happy to migrate to one of those pipelines and would love any assistance that OpenFF could provide in helping me do that.  :)

mattwthompson:
Fixed as part of #198

PR #196, opened 2020-12-23
Migrate CI to GitHub Actions

mattwthompson:
This is very much a work in progress, but has to be open since it's not a change that can be tested locally


mattwthompson:
Ah, I can't run this in a PR from a fork, so I'll need to set up in my fork first

PR #198, opened 2021-01-20
GitHub actions ci

leeping:
From here: https://github.com/mattwthompson/forcebalance/pull/1

This PR also fixes an issue with TINKER version parsing.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,128 @@
...
+#        wget https://raw.githubusercontent.com/leeping/forcebalance/master/tools/install-cctools.sh
+#        bash install-cctools.sh
+#        echo "Checking for Work Queue import; if successful, no message will be printed"
+#        python -c "import work_queue"
+#        export PATH="$GITHUB_WORKSPACE/opt/cctools/current/bin:$PATH"
+
+    - name: Install GROMACS
+      shell: bash -l {0}
+      run: |
+        # This will not install double precision, needs to be replaced with a fresh build
```
j-wags, line 73: Since it's easy to lose the context of these sorts of comments
```suggestion
        # TODO: As a result of the move from Travis to GitHub Actions, the CI no longer builds
        # GROMACS from source with double precision. This will need to be reverted so 
        # that the bromine test can return to its original strictness.
```

src/tinkerio.py
```diff
@@ -360,7 +360,7 @@ def calltinker(self, command, stdin=None, print_to_screen=False, print_command=F
                 if len(vw.split('.')) <= 2:
                     vn = float(vw)
                 else:
-                    vn = float(vw.split('.')[:2])
+                    vn = float(vw.split('.')[0]) + 0.1*float(vw.split('.')[1])
```
j-wags, line 363: I think this is a fix for the known bug that @leeping mentioned in the slack chat. @leeping, to ensure that there's a paper trail, could you edit the PR description to include something like
```
- [x] Fixes <description of issue here>
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,128 @@
...
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - macOS-latest
+          - ubuntu-latest
+        python-version:
+          - 3.6
```
j-wags, line 21:
```suggestion
          # TODO: Python 2.7 tests were lost during migration from Travis to Github Actions
          - 3.6
```

j-wags (approved):
Added a few suggestions. @leeping I see there's also a change for a different bug -- Could you document that in the PR body text?

I think this is in a good place to merge, but it'll be easy to lose track of the remaining to-dos, so three issues should be opened first:
* EITHER putting the OE license back, OR changing the little remaining functionality that requires it over to SDF (should be possible now that OFFTK supports partial-charges-in-SDF). I'd think that the latter is a better choice since it will remove the need for an OE license in CI and will let @leeping allow PRs from forks.
* Re-enabling Py27 tests
* Re-establishing double precision GROMACS builds

src/tinkerio.py
leeping: This issue was reported by a student in a private message but I don't think there's an issue corresponding to it.

leeping:
I've created an issue for all three items in https://github.com/leeping/forcebalance/issues/199 .  Thanks a lot for your help with the review! :)

PR #200, opened 2021-02-04
Correct a matrix for normal mode rearrangement

hyejang:
#### Description
In `vibration.py`, changed definition of `a` matrix for a correct normal mode rearrangement. 

#### Status
- [x]   Ready to go


leeping (approved):
Thanks and nice catch!

PR #201, opened 2021-02-10
Migrate to the new OpenFF toolkit namespace package

SimonBoothroyd:
## Description

This PR handles the migrations needed to use the `openff-toolkit` package (formerly named `openforcefield`)  which is now available on conda-forge. 

The only real difference between the `openforcefield` package previously used here and the new `openff-toolkit` package is all instances of `import openforcefield.XXX` and `from openforcefield.XXX` are now replaced by `import openff.toolkit.XXX` and `from openff.toolkit.XXX`

## Status

- [x] Ready to go


leeping (approved):

leeping:
Merging; thanks!

PR #202, opened 2021-02-11
Drude code for openmmio.py

jstoppelman:
I added the Drude code in openmmio.py. There were a few other small changes I made, such as readers for the CustomBond, ImproperTorsion and RBTorsionForce. Also, in vibration.py, I believe you'll run into an issue in vib_overlap if you have virtual sites or Drude particles in OpenMM and you don't specifically use masses for real atoms, so I included my fix for this.


leeping (approved):

leeping:
Thank you very much for this contribution! I think it looks great, merging it.

PR #203, opened 2021-02-11
Support Fitting SMIRNOFF 1-4 Scale Factors

SimonBoothroyd:
## Description

This PR adds support for fitting the 1-4 scale factors (both vdW and electrostatic) in SMIRNOFF based force fields. This mostly includes updating how SMIRNOFF parameters are cached in order to support non-quantity values and 'parameters' which do not include SMIRKS patterns.

## Status

- [ ] Ready to go


src/smirnoffio.py
```diff
@@ -145,40 +146,93 @@ def assign_openff_parameter(ff, new_value, pid):
...
+                )
+            )
+
+        # We can't use the caching approach when the parameter is a simple float,
+        # hence we set it directly here.
+        if isinstance(param_quantity, (float, int)):
+            setattr(parameter_container, value_name, new_value)
             param_quantity = None
         else:
-            # If the value name is a periodic attribute (say, k1) then we need to use
```
SimonBoothroyd, line 159: This code should no longer be needed as the toolkit supports accessing `k1`, `k2` etc directly now.

leeping:
Thanks for sharing, I think this looks great.  Could you add an example or modify an existing example where the scale factor is optimized?  I'd like to see what the string for the 1-4 scaling factor looks like.

SimonBoothroyd:
Sure! I've added an example in ff3f7e4

leeping (approved):

PR #206, opened 2021-03-12
Allow tinkerio to read TINKERPATH variable from environment

bdw2292:


leeping:
Thanks for making this PR. Is TINKERPATH the recommended environment variable name from the TINKER distribution or is it a custom name? Is your intention that the environment variable should override what is in the input file?

bdw2292:
Of course. I suppose it is a custom name. The only tinker wrapper I am
aware of that use something like this is poltype, which actually uses
TINKERDIR, however I wanted it to be more similar to the
forcebalance variable tinkerpath. The intention is to override the force
balance input with whatever is in the environment.

bdw2292:
It would not be an issue if I was only running on one node, however,
distributing jobs requires multiple environments potentially.

leeping:
I think that typically, the environment variable is more "invisible" whereas command line / input file options are more explicitly set by the user.  Could you switch the priority order so that the input file will override the environment variable?  You can still get the behavior you want by leaving tinkerpath blank in the input file.

src/tinkerio.py
```diff
@@ -314,14 +314,18 @@ def __init__(self, name="tinker", **kwargs):
...
+
     def setopts(self, **kwargs):
         
         """ Called by __init__ ; Set TINKER-specific options. """
 
         ## The directory containing TINKER executables (e.g. dynamic)
         if 'tinkerpath' in kwargs:
             self.tinkerpath = kwargs['tinkerpath']
-            if not os.path.exists(os.path.join(self.tinkerpath,"dynamic")):
+            if not os.path.exists(os.path.join(self.tinkerpath,"dynamic")) and 'dynamic' not in os.environ.keys():
```
leeping, line 328: Could you comment on this? Is "dynamic" supposed to be an environment variable too?

src/tinkerio.py
```diff
@@ -910,13 +919,13 @@ def molecular_dynamics(self, nsteps, timestep, temperature=None, pressure=None,
         if temperature is not None:
             md_defs["integrator"] = "stochastic"
         else:
-            md_defs["integrator"] = "beeman"
+            md_defs["integrator"] = "respa"
```
leeping, line 922: The default parameters are what's used if the user doesn't specify the option in the key file, so it doesn't need to be the most efficient, it only needs to be fail-safe.

Does the use of respa require any additional options in the key file, i.e. specifying the long and short timestep, or the subdivision of the long timestep to make the short timestep?

src/tinkerio.py
```diff
@@ -933,7 +942,7 @@ def molecular_dynamics(self, nsteps, timestep, temperature=None, pressure=None,
 
         eq_opts = deepcopy(md_opts)
         if self.pbc and temperature is not None and pressure is not None: 
-            eq_opts["integrator"] = "beeman"
+            eq_opts["integrator"] = "respa"
```
leeping, line 945: See above comment.

src/tinkerio.py
```diff
@@ -947,22 +956,37 @@ def molecular_dynamics(self, nsteps, timestep, temperature=None, pressure=None,
...
             if verbose: printcool("Running equilibration dynamics", color=0)
+            if 'OPENMM_CUDA_COMPILER' in os.environ.keys() and 'gas' not in self.name:
+                dynamickeyword='dynamic_omm'
+            else:
+                dynamickeyword='dynamic'
             if self.pbc and pressure is not None:
-                self.calltinker("dynamic %s -k %s-eq %i %f %f 4 %f %f" % (self.name, self.name, nequil, timestep, float(nsave*timestep)/1000, 
-                                                                          temperature, pressure), print_to_screen=verbose)
+                if 'OPENMM_CUDA_COMPILER' in os.environ.keys():
+                    self.calltinker(dynamickeyword+" %s -k %s-eq %i %f %f 4 %f %f N" % (self.name, self.name, nequil, timestep, float(nsave*timestep)/1000,temperature, pressure), print_to_screen=verbose)
```
leeping, line 965: Please add a comment regarding the extra "N" command line argument

src/tinkerio.py
```diff
@@ -947,22 +956,37 @@ def molecular_dynamics(self, nsteps, timestep, temperature=None, pressure=None,
...
             else:
-                self.calltinker("dynamic %s -k %s-eq %i %f %f 2 %f" % (self.name, self.name, nequil, timestep, float(nsave*timestep)/1000,
-                                                                       temperature), print_to_screen=verbose)
+                self.calltinker(dynamickeyword+" %s -k %s-eq %i %f %f 2 %f" % (self.name, self.name, nequil, timestep, float(nsave*timestep)/1000,temperature), print_to_screen=verbose)
             os.system("rm -f %s.arc" % (self.name))
 
         # Run production.
         if verbose: printcool("Running production dynamics", color=0)
         write_key("%s-md.key" % self.name, md_opts, "%s.key" % self.name, md_defs)
+        if 'OPENMM_CUDA_COMPILER' in os.environ.keys() and 'gas' not in self.name:
```
leeping, line 977: This seems to duplicate line 956, could the determination of dynamickeyword be moved "above" the "if nequil>0" block, so it doesn't need to be duplicated?

src/tinkerio.py
```diff
@@ -947,22 +956,37 @@ def molecular_dynamics(self, nsteps, timestep, temperature=None, pressure=None,
...
+             dynamickeyword='dynamic_omm'
+        else:
+             dynamickeyword='dynamic'
+
         if self.pbc and pressure is not None:
-            odyn = self.calltinker("dynamic %s -k %s-md %i %f %f 4 %f %f" % (self.name, self.name, nsteps, timestep, float(nsave*timestep/1000), 
-                                                                             temperature, pressure), print_to_screen=verbose)
+            if 'OPENMM_CUDA_COMPILER' in os.environ.keys():
```
leeping, line 983: Is there some way to simplify this so that it doesn't check for OPENMM_CUDA_COMPILER twice (once when it determines dynamickeyword and again when it decides how to call TINKER?)  Maybe you could make the N into a variable "dynamicsuffix" or something like that?

src/tinkerio.py
```diff
@@ -996,6 +1019,10 @@ def molecular_dynamics(self, nsteps, timestep, temperature=None, pressure=None,
         havekeys = set()
         first_shot = True
         for ln, line in enumerate(oanl):
+            if 'Polarization' in line:
```
leeping, line 1022: Can you please comment on why you want to skip parsing these lines out of analyze?

bdw2292:
In src/tinkerio.py
<https://github.com/leeping/forcebalance/pull/206#discussion_r599840990>:

>      def setopts(self, **kwargs):



         """ Called by __init__ ; Set TINKER-specific options. """



         ## The directory containing TINKER executables (e.g. dynamic)

         if 'tinkerpath' in kwargs:

             self.tinkerpath = kwargs['tinkerpath']

-            if not os.path.exists(os.path.join(self.tinkerpath,"dynamic")):

+            if not os.path.exists(os.path.join(self.tinkerpath,"dynamic"))
and 'dynamic' not in os.environ.keys():

Could you comment on this? Is "dynamic" supposed to be an environment
variable too?

Essentially, I had originally made an environment variable that would allow
dynamic_omm to look like dynamic so this program can run for GPU. I however
have changed this in the lastest commit (its possible I did not remove this
line), however now, there is a dynamickeyword that is dynamic_omm (the GPU
program name) if there is a keyword related to OPENMM/CUDA in the
environment (this keyword is necessary to be able to run).
------------------------------

In src/tinkerio.py
<https://github.com/leeping/forcebalance/pull/206#discussion_r599843389>:

> @@ -910,13 +919,13 @@ def molecular_dynamics(self, nsteps, timestep,
temperature=None, pressure=None,

         if temperature is not None:

             md_defs["integrator"] = "stochastic"

         else:

-            md_defs["integrator"] = "beeman"

+            md_defs["integrator"] = "respa"

The default parameters are what's used if the user doesn't specify the
option in the key file, so it doesn't need to be the most efficient, it
only needs to be fail-safe.

Does the use of respa require any additional options in the key file, i.e.
specifying the long and short timestep, or the subdivision of the long
timestep to make the short timestep?

Yes, I just had to change to respa, since Beeman does not exist for GPU
dynamic_omm. There are no additional options specific to that integrator.
However, I did need to add an extra “N” to a yes/no question for asking to
print out info from GPU for the dynamic_omm program (only if keyword for
OPENMM/CUDA is in environment) .
------------------------------

In src/tinkerio.py
<https://github.com/leeping/forcebalance/pull/206#discussion_r599843447>:

>              md_opts["thermostat"] = None

         # Periodic boundary conditions.

         if self.pbc:

             md_opts["vdw-correction"] = ''

             if temperature is not None and pressure is not None:

-                md_defs["integrator"] = "beeman"

+                md_defs["integrator"] = "respa"

See above comment.
------------------------------

In src/tinkerio.py
<https://github.com/leeping/forcebalance/pull/206#discussion_r599843549>:

> @@ -933,7 +942,7 @@ def molecular_dynamics(self, nsteps, timestep,
temperature=None, pressure=None,



         eq_opts = deepcopy(md_opts)

         if self.pbc and temperature is not None and pressure is not None:

-            eq_opts["integrator"] = "beeman"

+            eq_opts["integrator"] = "respa"

See above comment.
------------------------------

In src/tinkerio.py
<https://github.com/leeping/forcebalance/pull/206#discussion_r599843997>:

>              if self.pbc and pressure is not None:

-                self.calltinker("dynamic %s -k %s-eq %i %f %f 4 %f %f" % (
self.name, self.name, nequil, timestep, float(nsave*timestep)/1000,

-
temperature, pressure), print_to_screen=verbose)

+                if 'OPENMM_CUDA_COMPILER' in os.environ.keys():

+                    self.calltinker(dynamickeyword+" %s -k %s-eq %i %f %f
4 %f %f N" % (self.name, self.name, nequil, timestep,
float(nsave*timestep)/1000,temperature, pressure), print_to_screen=verbose)

Please add a comment regarding the extra "N" command line argument

“However, I did need to add an extra “N” to a yes/no question for asking to
print out info from GPU for the dynamic_omm program (only if keyword for
OPENMM/CUDA is in environment).” Just extra item that dynamic_omm has over
dynamic.


------------------------------

In src/tinkerio.py
<https://github.com/leeping/forcebalance/pull/206#discussion_r599844852>:

>              os.system("rm -f %s.arc" % (self.name))



         # Run production.

         if verbose: printcool("Running production dynamics", color=0)

         write_key("%s-md.key" % self.name, md_opts, "%s.key" % self.name,
md_defs)

+        if 'OPENMM_CUDA_COMPILER' in os.environ.keys() and 'gas' not in
self.name:

This seems to duplicate line 956, could the determination of dynamickeyword
be moved "above" the "if nequil>0" block, so it doesn't need to be
duplicated?

Ah okay some redundancy, but it works 😊
------------------------------

In src/tinkerio.py
<https://github.com/leeping/forcebalance/pull/206#discussion_r599845780>:

>          if self.pbc and pressure is not None:

-            odyn = self.calltinker("dynamic %s -k %s-md %i %f %f 4 %f %f"
% (self.name, self.name, nsteps, timestep, float(nsave*timestep/1000),

-
temperature, pressure), print_to_screen=verbose)

+            if 'OPENMM_CUDA_COMPILER' in os.environ.keys():

Is there some way to simplify this so that it doesn't check for
OPENMM_CUDA_COMPILER twice (once when it determines dynamickeyword and
again when it decides how to call TINKER?) Maybe you could make the N into
a variable "dynamicsuffix" or something like that?

Yes this can be made to look “prettier” and more efficient.
------------------------------

In src/tinkerio.py
<https://github.com/leeping/forcebalance/pull/206#discussion_r599846532>:

> @@ -996,6 +1019,10 @@ def molecular_dynamics(self, nsteps, timestep,
temperature=None, pressure=None,

         havekeys = set()

         first_shot = True

         for ln, line in enumerate(oanl):

+            if 'Polarization' in line:

Can you please comment on why you want to skip parsing these lines out of
analyze?

Basically, new analyze has a line that says Polarization MUTAL or
Polarization THOLE, but program is trying to parse energy from line that
has Polarization energy value. So just preventing this from crashing
program with new analyze.

leeping:
@bdw2292 I'm reviewing all the open PRs and this one is looking pretty huge at this point, and it contains lots of code. I do want to include your features into the master branch, but I'm not sure I can understand all your modifications on my own. Could we schedule a meeting sometime to go over what you wrote? You may feel free to ask your PI to join as well.

Issue #209, opened 2021-03-22
AttributeError: 'ForceField' object has no attribute '_forces'


pavankum:
[](url)
[forcebalance_error.tar.gz](https://github.com/leeping/forcebalance/files/6184067/forcebalance_error.tar.gz)

Hi Lee-ping, 

I am getting an attribute error while trying to do a fit with the new release. I attached a minimal example along with the conda-env I used. Please take a look when you have a moment. 

Thank you!

leeping:
Hmm .. Could you change this line and let me know if it resolves the issue?

openmmio.py line 845
From this:
```
                elif any(['Drude' in f.__class__.__name__ for f in self.forcefield._forces]): integrator = DrudeLangevinIntegrator(temperature*kelvin, collision/picosecond, 1*kelvin, collision/picosecond, 0.1*femtoseconds)
```
To this:
```
                elif hasattr(self.forcefield, '_forces') and any(['Drude' in f.__class__.__name__ for f in self.forcefield._forces]): 
                    integrator = DrudeLangevinIntegrator(temperature*kelvin, collision/picosecond, 1*kelvin, collision/picosecond, 0.1*femtoseconds)
```

leeping:
See https://github.com/leeping/forcebalance/pull/210

PR #210, opened 2021-03-22
Use OpenMM.System instead of OpenMM.Forcefield to check for Drude force

leeping:
Resolves https://github.com/leeping/forcebalance/issues/209 .

The problem was that the `SMIRNOFF` class calls `OpenMM.create_simulation()`, and the recently added Drude feature was checking for `OpenMM.forcefield._forces`, which is something not possessed by `SMIRNOFF.forcefield`.  The issue is fixed by checking for the presence of Drude forces using `OpenMM.system.getForces()` instead.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,7 +71,7 @@ jobs:
       shell: bash -l {0}
       run: |
         # This will not install double precision, needs to be replaced with a fresh build
-        conda install gromacs=2020.5 -c bioconda -c conda-forge -y
+        conda install gromacs=4.6.5 -c bioconda -c conda-forge -y
 
     - name: Install Tinker
       run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -71,7 +71,7 @@ jobs:
       shell: bash -l {0}
       run: |
         # This will not install double precision, needs to be replaced with a fresh build
-        conda install gromacs -c bioconda -c conda-forge -y
+        conda install gromacs=4.6.5 -c bioconda -c conda-forge -y
```
mattwthompson, line 74: What about trying 2019 or 2018? Versions 2019.1 and 2018.6 were previously being pulled down, i.e. these CI logs from a week ago: https://github.com/leeping/forcebalance/runs/2115435511

```suggestion
        conda install gromacs=2019.1 -c bioconda -c conda-forge -y
```

mattwthompson:
Looks like that fixed it!

PR #214, opened 2021-04-01
Initial support for fitting SMIRNOFF virtual sites

SimonBoothroyd:
## Description

This PR aims to add initial support for fitting virtual sites which are defined in a SMIRNOFF based force field. A new example `027_smirnoff_tip4p` is added as a test of the new support only, and not meant to be scientifically sensible much like `026_drude_water`.

## Status

- [ ] Ready to go


src/smirnoffio.py
```diff
@@ -383,12 +434,21 @@ def prepare(self, pbc=False, mmopts={}, **kwargs):
...
-            self.xyz_omms.append((position, box_omm))
+            self.xyz_omms.append((xyz_omm, box_omm))
+
+        # used in create_simulation()
+        openmm_topology = SMIRNOFF._openff_to_openmm_topology(openff_topology)
+        openmm_positions = (
+            self.pdb.positions.value_in_unit(angstrom) +
+            # Add placeholder positions for an v-sites.
+            [Vec3(0.0, 0.0, 0.0)] * openff_topology.n_topology_virtual_sites
+        ) * angstrom
```
leeping, line 445: Also it looks like the positions of virtual sites are always appended to the end of the entire structure. Does this ever lead to issues?

src/smirnoffio.py
```diff
@@ -433,13 +495,33 @@ def update_simulation(self, **kwargs):
         #         delattr(self, 'simulation')
         # self.vsprm = vsprm.copy()
 
+        if openff_topology.n_topology_virtual_sites > 0:
+            # For now always assume that the v-sites have changed. This is currently
+            # needed as the FB checks don't support the ``LocalCoordinatesSite`` based
+            # virtual sites that OpenFF uses.
+            if hasattr(self, 'simulation'):
+                delattr(self, 'simulation')
+
```
leeping, line 504: I'm wondering why you had to do that? Is there no way to update the virtual site parameters without deleting the Simulation object? In practice does it affect performance?

src/smirnoffio.py
```diff
@@ -305,6 +321,36 @@ def readsrc(self, **kwargs):
         # Store a separate copy of the molecule for reference restraint positions.
         self.ref_mol = deepcopy(self.mol)
 
+    @staticmethod
```
leeping, line 324: I'm fine with having this in FB but shouldn't this be an OpenFF feature? Is this here as a temp-fix because the feature isn't due to be released in OpenFF until some point in the future?

src/smirnoffio.py
```diff
@@ -160,7 +166,17 @@ def assign_openff_parameter(ff, new_value, pid):
         else:
             (handler_name, tag_name, value_name, smirks) = pid.split('/')
 
+            from openff.toolkit.typing.engines.smirnoff import ParameterList
```
leeping, line 169: Is ParameterList a new class? It would be nice to have a few lines of comments here because I am confused by this and line 176-178.

leeping (commented):
Thank you for the examples! I think everything looks good but a few explanations would be great prior to merge. I will do my best to not delay things any further.

src/smirnoffio.py
SimonBoothroyd: I think the OpenFF toolkit made the decision to add virtual sites at the end of a topology so that the indices of atoms were preserved. @j-wags can probably comment on that a bit more though. As far I can tell this doesn't cause issues, although it is a little more awkward to work with.

src/smirnoffio.py
SimonBoothroyd: I believe I did this as the `PrepareVirtualSites` function did not yet support `LocalCoordinatesSite` which the OpenFF toolkit seems to use exclusively. Once #215 is merged however it would seem that this might not be needed.

src/smirnoffio.py
SimonBoothroyd: > Is this here as a temp-fix because the feature isn't due to be released in OpenFF until some point in the future?

Unfortunately this is currently the case. @j-wags might have more information on a timeline for this however?

src/smirnoffio.py
SimonBoothroyd: Ah good catch - this was needed to fix https://github.com/openforcefield/openff-toolkit/issues/884 however now that the fix is merged upstream we should be safe to remove this!

leeping (approved):

PR #215, opened 2021-05-04
`LocalCoordinatesSite` support for OpenMM

jthorton:
This PR implements a virtual site position calculation function for `LocalCoordinatesSite` type sites which currently are not supported [here](https://github.com/leeping/forcebalance/blob/60eeb84d10c87dbfa07aff35248ca97eea8fecbd/src/openmmio.py#L126) and raise an error. A test is also included to ensure that the positions calculated agree with those given by OpenMM.


leeping:
Thank you! This looks very reasonable, merging it.

leeping (approved):

leeping (approved):

PR #219, opened 2021-05-12
Add option to delete micro outputs

SimonBoothroyd:
## Description

When fitting against a large number of targets the number of `micro` directories created in `output.tmp`, and the number of files they contain, can grow quite large. This is especially true when the full optimisation takes many iterations >=20 to converge, and each individual iteration can require ~20-40 micro iterations before selecting a step size.

This PR adds an option to delete these `micro` directories after they are finished with, thus reducing the total number of files produced by an optimization. By default they are retained to preserve previous behaviour.

### DIFF
--- a/src/optimizer.py
+++ b/src/optimizer.py
@@ -10,6 +10,7 @@ contained inside.
 from __future__ import division
 from __future__ import print_function
 
+import shutil
 from builtins import str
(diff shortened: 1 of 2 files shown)


leeping (approved):
This is a great feature - thanks a lot!

PR #224, opened 2021-05-28
Switching to lp_load

bieniekmateusz:
The direct pickle.load is trying to read a compressed file.

### DIFF
--- a/src/optimizer.py
+++ b/src/optimizer.py
@@ -21,7 +21,7 @@ from numpy.linalg import multi_dot
 from copy import deepcopy
 import forcebalance
 from forcebalance.parser import parse_inputs
-from forcebalance.nifty import col, flat, row, printcool, printcool_dictionary, pvec1d, pmat2d, warn_press_key, invert_svd, wopen, bak, est124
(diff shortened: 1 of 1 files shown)


bieniekmateusz:
Sorry, I think that's my mistake. I'll check again why unpickling does not work for me.

bieniekmateusz:
That's correct. When I restart my master workload server, it throws this error:
```
  File "/lustre/scafellpike/local/HT03119/mjm06/mxb57-mjm06/software/anaconda3/envs/fb/lib/python3.7/site-packages/forcebalance/optimizer.py", line 1564, in readchk
    self.chk = pickle.load(open(absfnm))
  File "/lustre/scafellpike/local/HT03119/mjm06/mxb57-mjm06/software/anaconda3/envs/fb/lib/python3.7/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte
```
However, using lp_load works just fine and extracts the variable. This is on the `restart.p` file. If you think someone's wrong, please let me know!

Thanks, Mat

leeping:
Thank you, and sorry for the late review!  I'm merging your fix. :)

leeping (approved):

PR #228, opened 2021-06-30
Fix #197 and Hessian import errors

### LINKED ISSUE #197: smirnoff_hack SMARTS caching does not work if two instances of a molecule are in different order
Cross-posting this from https://github.com/openforcefield/openff-toolkit/issues/812 to just make this repo aware of this small yet subtle issue.

Essentially, two molecules with different atom orderings will hit the same cache line, and the returned indices for the second molecule will be that of the first, which is incorrect. The underlying cause for this is that the hack uses the molecule hash to establish uniqueness, but currently the toolkit is just hashing the canonical smiles, which is the same for both molecules.

SimonBoothroyd:
## Description

This PR fixes #197 by making sure the `smirnoff_hack.py` uses a hash that takes the order of molecule into account. It also introduces a simple mechanism to disable the caching provided by `smirnoff_hack.py` - this is useful to temporarily disable the caching if changes in an upstream dependency (mainly the toolkit) breaks the implemented methods and a new `forcebalance` release has not yet been made to support the changes.

## Status

- [ ] Ready to go


src/smirnoff_hack.py
```diff
@@ -1,120 +1,158 @@
...
+    # SMIRNOFF_FORCE_FIELD_CACHE = {}
+    # def getForceField(*ffpaths):
+    #     hasher = hashlib.md5()
+    #     for path in ffpaths:
+    #         with open(path, 'rb') as f:
+    #             hasher.update(f.read())
+    #     cache_key = hasher.hexdigest()
+    #     if cache_key not in SMIRNOFF_FORCE_FIELD_CACHE:
+    #         SMIRNOFF_FORCE_FIELD_CACHE[cache_key] = ForceField(*ffpaths, allow_cosmetic_attributes=True)
+    #     return SMIRNOFF_FORCE_FIELD_CACHE[cache_key]
```
SimonBoothroyd, line 158: None of this code has changed - it has just been indented by 4 spaces to fall under the `if` statement.

SimonBoothroyd, line 158: None of this code has changed - it has just been indented by 4 spaces to fall under the if statement.

leeping (approved):

PR #231, opened 2021-08-10
Smirnoff Hack update

jthorton:
This PR updates the Smirnoff hack file to work with the newest toolkit release, introducing a new argument to the partial charge assignment methods that normalize the net charge.

### DIFF
--- a/src/smirnoff_hack.py
+++ b/src/smirnoff_hack.py
@@ -33,6 +33,12 @@ def hash_molecule(molecule):
     return cmiles
 
 
+def hash_molecule_args_and_kwargs(molecule, *args, **kwargs):
+    arguments = [str(arg) for arg in args]
+    keywords = [str(keyword) for keyword in kwargs.values()]
+    return hash((hash_molecule(molecule), *arguments, *keywords))
+
+
 if _SHOULD_CACHE:
 
     print(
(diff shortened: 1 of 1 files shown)


j-wags:
Thanks for opening this, @jthorton. I keep forgetting about this file -- this is such an interesting use case, and I never really consider that an API _extension_ could break something. 

In this case, we may need a try/except to support both the old (<=0.9.2) API and the new (>=0.10.0) API. Or we could hash based on the method signature and not hard-code the actual parameters at all (though I'm not sure whether this is safe)

```python
sig = inspect.signature(AmberToolsToolkitWrapper.assign_partial_charges)
print(sig)
print(hash(sig))
```

> (self, molecule, partial_charge_method=None, use_conformers=None, strict_n_conformers=False, normalize_partial_charges=True, _cls=None)
> 1623433902460278560

jthorton:
@j-wags I think if we could future proof this file that would be great - how to do this however I am not so sure. I think taking `*args/**kwargs` would be great and I need to look through forcebalance to double-check this but I believe FB never changes any default settings for the charge method so maybe we could just hash based on the molecule? Tagging @SimonBoothroyd in case recharge would be affected.

SimonBoothroyd:
+1 for `*args / **kwargs`

j-wags (approved):
I've tested this against openFF toolkit 0.9.2 (before the API change) and 0.10.0 (after the API change), and `test_smirnoff_hack` passes for both. 

The hash code also looks good to me. There may be a slight performance hit for hashing the string of a large numpy array if `use_conformers` is not `None` (instead of hashing the numpy object directly). However the human-minutes it would take to fix that exceed the additional computational cost that this will likely consume for the remaining lifetime of the universe. So no objections there. 

I'll let @leeping give final approval, but this looks good from the OFF Toolkit side :-)

PR #237, opened 2021-08-25
[Evaluator] Support gradients of handler attributes

SimonBoothroyd:
## Description

This PR updates the OpenFF Evaluator target so that it can correctly handle optimising handler attributes (e.g. the 1-4 scaling factor or the alpha, beta parameters of the [double exponential handler](https://github.com/openforcefield/smirnoff-plugins/pull/13))

PR #239, opened 2021-09-17
Include conditional imports for openmm 7.6 compatibility

j-wags:
CI is failing due to the openmm namespace change in OpenMM version 7.6. This PR will support importing from either namespace. I've tested this locally using `openmm=7.6.0` and `openmm=7.5.1`. I will merge this when CI passes.


j-wags:
Never mind -- I'll merge these changes in with #237 to avoid the need for separate approval.

PR #240, opened 2021-09-17
Update all simtk imports to also be forward-compatible with OpenMM >7.6, and keep soft imports soft

j-wags:
My changes in #237 made ForceBalance work with EITHER openmm >=7.6 OR <7.6. However, I didn't take into consideration builds with no openmm at all. Since openmmio is imported even in the absense of an installed openmm package, the imports need to allow parsing of the file regardless of whether openmm is installed.

Since there are many variables here ({hard vs. soft imports}, {openmm 7.5 vs. 7.6 vs. no openmm}, {tests vs. studies}...) we should be pretty thorough in our checks. I propose the following:

- [x] Ensure vanilla CI passes and doesn't skip openmmio tests ([done](https://github.com/leeping/forcebalance/runs/3637497811#step:12:75))
- [x] Ensure CI that's forced to use OpenMM 7.5 (with the old namespace) passes and doesn't skip openmmio tests ([done](https://github.com/leeping/forcebalance/runs/3637495901#step:12:75))
- [ ] Ensure that studies still run with OpenMM 7.5
- [ ] Ensure that studies still run with OpenMM 7.6
- [ ] Ensure that non-openMM-requiring ForceBalance functionality is available without installing OpenMM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,6 +107,7 @@ jobs:
         # python setup.py install
         python -m pip install --no-deps .
         python -c "import forcebalance; print(forcebalance.__version__)"
+        conda remove --force openmm
 
     - name: Run water study
       shell: bash -l {0}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,7 +107,6 @@ jobs:
         # python setup.py install
         python -m pip install --no-deps .
         python -c "import forcebalance; print(forcebalance.__version__)"
-        conda remove --force openmm
 
     - name: Run water study
       shell: bash -l {0}
--- a/src/data/md_ism_hfe.py
+++ b/src/data/md_ism_hfe.py
@@ -89,9 +89,14 @@ engname = TgtOpts['engname']
 # Import modules and create the correct Engine object.
 if engname == "openmm":
     try:
-        from simtk.unit import *
-        from simtk.openmm import *
-        from simtk.openmm.app import *
+        try:
+            from openmm.unit import *
+            from openmm import *
+            from openmm.app import *
+        except ImportError:
+            from simtk.unit import *
+            from simtk.openmm import *
+            from simtk.openmm.app import *
     except:
         traceback.print_exc()
         raise Exception("Cannot import OpenMM modules")
--- a/src/data/npt.py
+++ b/src/data/npt.py
@@ -52,9 +52,15 @@ engname          = args.engine.lower()     # Name of the engine
 
 if engname == "openmm":
     try:
-        from simtk.unit import *
-        from simtk.openmm import *
-        from simtk.openmm.app import *
+        try:
+            from openmm.unit import *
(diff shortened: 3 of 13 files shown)


leeping:
I'm still seeing too many skips on the CI tests. There should only be two intended skips in this list of twelve (those highlighted):

```src/tests/test_engine.py::TestAmber99SB::test_energy_force SKIPPED (...) [  5%]
src/tests/test_engine.py::TestAmber99SB::test_optimized_geometries SKIPPED [  6%]
src/tests/test_engine.py::TestAmber99SB::test_interaction_energies SKIPPED [  7%]
src/tests/test_engine.py::TestAmber99SB::test_multipole_moments SKIPPED  [  9%]
src/tests/test_engine.py::TestAmber99SB::test_multipole_moments_optimized SKIPPED [ 10%]
src/tests/test_engine.py::TestAmber99SB::test_normal_modes SKIPPED (...) [ 11%]
src/tests/test_engine.py::TestAmber99SB::test_normal_modes_optimized SKIPPED [ 12%]
src/tests/test_engine.py::TestAmoebaWater6::test_energy_force PASSED     [ 14%]
**src/tests/test_engine.py::TestAmoebaWater6::test_energy_rmsd SKIPPED     [ 15%]**
src/tests/test_engine.py::TestAmoebaWater6::test_interaction_energy PASSED [ 16%]
src/tests/test_engine.py::TestAmoebaWater6::test_multipole_moments PASSED [ 18%]
**src/tests/test_engine.py::TestAmoebaWater6::test_multipole_moments_optimized SKIPPED [ 19%]**
```

@mattwthompson Could you comment on whether we skipped extra tests when we migrated away from Travis CI to the current system?  If not, then I think these skips are symptomatic of the environment not being correctly loaded.

j-wags:
I think those skips were there for the 1.9.0 release as well -- Comparing the CI results of [the 1.9.0 release from 3 months ago](https://github.com/leeping/forcebalance/runs/3105790080?check_suite_focus=true#step:12:20) to [the CI run from the 1.9.1 release last Friday](https://github.com/leeping/forcebalance/runs/3636434633?check_suite_focus=true#step:12:20):

```python
>>> print(before.count('PASSED'), before.count('SKIPPED'))
63 14
>>> print(after.count('PASSED'), after.count('SKIPPED'))
63 14
```

leeping:
Got it, thanks. We'll kick that particular can down the road. :)

leeping (approved):
Looks good to me!

mattwthompson:
I'm late to the party but it looks like things are in order - at least no more broken than they were before. I fully agree with what's already been set in motion - merging this and trying to un-skip those in a separate PR.

My memory is hazy but it makes sense to me that a few extra skips would have slipped in there during the CI migration. Whether or not I noted it then and didn't come back to fix it, it's really easy for the number of skipped tests to change since it requires actually looking through the logs. I'm often guilty of this, and that's probably what happened there.

PR #241, opened 2021-09-20
Code updates to enable TestLiquid_OpenMM

leeping:
Enables two OpenMM tests that were unreasonably turned off at an earlier date.

### DIFF
--- a/src/liquid.py
+++ b/src/liquid.py
@@ -253,7 +253,7 @@ class Liquid(Target):
 
     def read_data(self):
         # Read the 'data.csv' file. The file should contain guidelines.
-        with open(os.path.join(self.tgtdir,'data.csv'),'rU') as f: R0 = list(csv.reader(f))
+        with open(os.path.join(self.tgtdir,'data.csv'),'r') as f: R0 = list(csv.reader(f))
         # All comments are erased.
         R1 = [[sub('#.*$','',word) for word in line] for line in R0 if len(line[0]) > 0 and line[0][0] != "#"]
         # All empty lines are deleted and words are converted to lowercase.
@@ -727,7 +727,6 @@ class Liquid(Target):
         @return property_results
 
         """
-
         unpack = lp_load('forcebalance.p')
         mvals1 = unpack[1]
         if len(mvals) > 0 and (np.max(np.abs(mvals1 - mvals)) > 1e-3):
--- a/src/target.py
+++ b/src/target.py
@@ -598,7 +598,7 @@ class Target(with_metaclass(abc.ABCMeta, forcebalance.BaseClass)):
     def submit_jobs(self, mvals, AGrad=False, AHess=False):
         return
 
-    def stage(self, mvals, AGrad=False, AHess=False, customdir=None, firstIteration=False):
+    def stage(self, mvals, AGrad=False, AHess=False, use_iterdir=True, customdir=None, firstIteration=False):
         """
 
         Stages the directory for the target, and then launches Work Queue processes if any.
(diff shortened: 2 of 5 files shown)


src/tests/test_openmmio.py
```diff
@@ -12,26 +12,28 @@
...
         super(TestLiquid_OpenMM, self).setup_method(method)
+        self.check_grad_fd = False
         # settings specific to this target
         self.options.update({
                 'jobtype': 'NEWTON',
                 'forcefield': ['dms.xml']})
 
         self.tgt_opt.update({'type':'LIQUID_OPENMM',
-            'name':'dms-liquid'})
+            'name':'dms-liquid', 'liquid_md_steps':500})
```
j-wags: (not blocking) I'm just curious - is this the change that enables the liquid tests?

j-wags (approved):
I do see the two new tests running and passing now, and no other change in test output. And the changes look good, though I'm not totally sure which of them is responsible for enabling the new tests :-)

mattwthompson:
(Looks like Jeff beat me to it)

Looking at nothing more than the number of passed versus skipped test, this seems to do what's advertised: +2 passing and -2 passing.

Copying from the logs (of 84cb02693ab5d03013f2ded61631e69909733440)
```
=========== 65 passed, 12 skipped, 23 warnings in 1091.40s (0:18:11) ===========
```

compared to 63 and 14 reported earlier: https://github.com/leeping/forcebalance/pull/240#issuecomment-923153176

src/tests/test_openmmio.py
```diff
@@ -12,26 +12,28 @@
...
         self.tgt_opt.update({'type':'LIQUID_OPENMM',
-            'name':'dms-liquid'})
+            'name':'dms-liquid', 'liquid_eq_steps':100, 'liquid_md_steps':200, 'gas_eq_steps':100, 'gas_md_steps':200})
 
         self.ff = forcebalance.forcefield.FF(self.options)
 
         self.ffname = self.options['forcefield'][0][:-3]
         self.filetype = self.options['forcefield'][0][-3:]
-        self.mvals = [.5]*self.ff.np
+        self.mvals = np.array([.5]*self.ff.np)
```
leeping, line 32: @j-wags This was one of the changes; at a different point in the code, it tried to subtract two mvals and encountered a problem when mvals was a list.

src/tests/test_target.py
```diff
@@ -17,6 +17,7 @@ def setup_method(self, method):
         self.tgt_opt = forcebalance.parser.tgt_opts_defaults.copy()
         self.ff = None  # Forcefield this target is fitting
         self.options.update({'root': os.path.join(os.getcwd(), 'files')})
+        self.check_grad_fd = True # Whether to check gradient vs. finite difference. Set to False for liquid targets.
```
leeping, line 20: @j-wags This was one of the changes; comparing analytic grad. to finite diff. would automatically fail for liquid properties because of the stochastic / statistical nature.

src/target.py
```diff
@@ -614,7 +614,7 @@ def stage(self, mvals, AGrad=False, AHess=False, customdir=None, firstIteration=
         cwd = os.getcwd()
 
         absgetdir = os.path.join(self.root,self.tempdir)
-        if Counter() is not None:
+        if use_iterdir and Counter() is not None:
```
leeping, line 617: @j-wags This was one of the changes; the `stage()` function automatically goes into an "iteration" folder but that is not where the `get()` function looks for the data in the test, so I had to add an extra keyword argument.

src/liquid.py
```diff
@@ -253,7 +253,7 @@ def prepare_temp_directory(self):
 
     def read_data(self):
         # Read the 'data.csv' file. The file should contain guidelines.
-        with open(os.path.join(self.tgtdir,'data.csv'),'rU') as f: R0 = list(csv.reader(f))
+        with open(os.path.join(self.tgtdir,'data.csv'),'r') as f: R0 = list(csv.reader(f))
```
leeping, line 256: @j-wags This should take care of a number of deprecation warnings.

j-wags:
Ah, thanks for the explanations. These changes all make sense to me now.

PR #243, opened 2021-09-24
Final changes for v1.9.2 release

leeping:
- Enabled several tests that were being skipped because package names were incorrect.
- Changed some dependencies from hard to soft, so that tests are skipped instead of error / fail in an OpenMM-less environment.
- Skip printing out of SMIRNOFF caching because it was being printed out even when it's not relevant.
- Changed `'rU'` to `'r'` in some `open()` statements to avoid deprecation warning.
- Changed `logger.warn()` to `logger.warning()` to avoid deprecation warning.
- Further increase tolerance of liquid bromine test.
- Updated version number to v1.9.2.


src/tests/test_system.py
```diff
@@ -181,7 +181,7 @@ def test_thermo_bromine_study(self):
 
 class TestEvaluatorBromineStudy(ForceBalanceSystemTest):
     def setup_method(self, method):
-        pytest.importorskip("evaluator")
+        pytest.importorskip("openff.evaluator")
```
j-wags, line 184: (not blocking) This is confusing. The previous import test should have always failed, so this test should have always skipped, but I was seeing this test failure on my computer before you made this change. So this may be "smoke" for something deeper, but it's not blocking because 1) my memory is bad and I could be mistaken and 2) we fixed the study by running it explicitly so we know that it works.

src/tests/__init__.py
```diff
@@ -27,3 +27,21 @@ def setup_method(self, method):
...
+        except ImportError:
+            # Try importing openmm using <7.6 namespace
+            import simtk.openmm as mm
+            from simtk.openmm import app
+            from simtk import unit
+        return True
+    except ImportError:
+        # If OpenMM classes cannot be imported, then set this flag 
+        # so the testing classes/functions can use to skip.
+        return False
```
j-wags, line 47: (not blocking) Clever. I like this.

j-wags (approved):
This looks great and should be good to go. I left two comments, neither is blocking. Thanks!

src/tests/test_system.py
leeping: That's definitely strange. It should be fixed now, and I don't see why the test would have failed unless your env had the old package name.

src/tests/__init__.py
leeping: Thanks! :)

j-wags:
Oh, and sorry to add another delay -- Let me do a dry run of the conda packaging test that caught the problem with the 1.9.1 release. I think it should overlap with the "no openmm" test that Lee Ping did locally, but it'll be good to ensure that we don't have to redo another release. I should have an answer in ~20 minutes.

j-wags:
Packing tests are showing green. I approve this merge and the release!

PR #248, opened 2022-03-01
Update smirnoff_hack.py for OpenFF Toolkit 0.10.3

mattwthompson:
We released a new version of the toolkit yesterday that broke `smirnoff_hack` by nature of adding a named argument to `generate_conformers` - fixing it should be a matter of string the new argument through the appropriate functions.

### DIFF
--- a/src/smirnoff_hack.py
+++ b/src/smirnoff_hack.py
@@ -128,7 +128,7 @@ if _SHOULD_CACHE:
     def oe_cached_generate_conformers(self, molecule, n_conformers=1, rms_cutoff=None, clear_existing=True, make_carboxylic_acids_cis=False):
         cache_key = hash((hash_molecule(molecule), n_conformers, str(rms_cutoff), clear_existing, make_carboxylic_acids_cis))
         if cache_key not in OE_TOOLKIT_CACHE_molecule_conformers:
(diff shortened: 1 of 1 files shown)


mattwthompson:
CI seems to be failing for other reasons (pip and/or other deps changing). I might be able to have a look later.

j-wags (commented):
Ahh gosh, sorry, I keep forgetting about this file. Yes, the OFF Toolkit 0.10.3 release yesterday added a kwarg to the `generate_conformers` method signature which broke the caching logic here. I'm sorry for the inconvenience @leeping. 

This code change looks correct to me. @leeping - ForceBalance's CI hasn't run since the 0.10.3 OFF toolkit release, so I can't show you the problem, but if we run CI on `master` right now it should fail because of this problem. 

Apologies again, I think our ask is both for this PR and for a ForceBalance release. There may be other issues happening too with other deps and we could discuss how to handle those as well.

mattwthompson:
I've updated this to be self-contained to the changes associated with the toolkit. I opened #249 in an effort to fix other failing tests. Hopefully that can be accomplished before this one.

j-wags:
The approach from #231 (switching from explicit listing of args and kwargs to using `*args` and `**kwargs`) would probably be a more future-proof way of handling this. 

In the long run it'd be great to fully move the caching out of forcebalance and into the toolkit. Maybe we could coordinate with @leeping on that so we can synchronize making that change and a subsequent release in the future? Then we could stabilize the FB interface and not cause this kind of problem in the future.

mattwthompson:
Agreed, I've opened #251 with a smaller set of changes

PR #249, opened 2022-03-01
Update CI, fix compatibility with OpenMM 7.7

### LINKED ISSUE #246: Compatibility with OpenMM-7.7.0
OpenMM 7.7 introduced a lot of argument checks in functions which are causing some tests to fail which involve liquid simulations see [here](https://github.com/leeping/forcebalance/runs/4970177002?check_suite_focus=true#step:12:393) for an example.

mattwthompson:
Resolves #246

### DIFF
--- a/src/openmmio.py
+++ b/src/openmmio.py
@@ -980,6 +980,8 @@ class OpenMM(Engine):
         #printcool_dictionary(self.mmopts, title="Creating/updating simulation in engine %s with system settings:" % (self.name))
         # for b in list(self.mod.topology.bonds()):
         #     print b[0].index, b[1].index
+        self.mmopts['useSwitchingFunction'] = False
+        self.mmopts['switchingDistance'] = 0.0 * self.mmopts['switchingDistance']
         self.system = self.forcefield.createSystem(self.mod.topology, **self.mmopts)
         self.vsinfo = PrepareVirtualSites(self.system)
         self.nbcharges = np.zeros(self.system.getNumParticles())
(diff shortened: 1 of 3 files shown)


src/openmmio.py
```diff
@@ -980,7 +980,12 @@ def update_simulation(self, **kwargs):
         #printcool_dictionary(self.mmopts, title="Creating/updating simulation in engine %s with system settings:" % (self.name))
         # for b in list(self.mod.topology.bonds()):
         #     print b[0].index, b[1].index
-        self.system = self.forcefield.createSystem(self.mod.topology, **self.mmopts)
+        try:
+            self.system = self.forcefield.createSystem(self.mod.topology, **self.mmopts)
+        except ValueError:
```
j-wags, line 985:
```suggestion
        # This try/except block catches a failure case introduced by the release of openmm 7.7
        # where a ValueError would be raised if createSystem was given an unused kwarg.
        # Now, when that error occurs, we remove the unused kwargs from mmopts.
        # More info at https://github.com/leeping/forcebalance/issues/246
        except ValueError as e:
            if 'useSwitchingFunction' not in str(e):
                raise e
```

j-wags (approved):
Looks good. Let's unpin openmm and openff-toolkit so that the python3.6 runs can resolve their environments. 

Also, let's wait on merging for one day (until noon PT on Mar 8), in case @leeping has any strong objection to the changes in `openmmio.py`. 

@leeping - The changes in `openmmio.py` are the minimal changes that we can put in to get ForceBalance working again, since OpenMM 7.7 now raises an error for unused kwargs, and in some cases where the OpenMM ForceField is making a PME system, it can't use a switching function. Alternatively we could have removed [the lines that request use of a switching function](https://github.com/leeping/forcebalance/blob/a8659b20979c405f192d589e5bdbb606dc6ea641/src/openmmio.py#L810-L811), but I'm not sure that they would fail in every possible case, so we tried to tailor the changes carefully to only disable switching if we KNOW that an error would be raised.

PR #251, opened 2022-03-02
Cache generate_conformers with hash_molecule_args_and_kwargs approach

mattwthompson:
This is a simpler approach than #248, as @j-wags pointed out.


j-wags (approved):
This looks good to me! Let's wait to merge #249 before we merge this, so that we can have the test matrix totally green when this goes in.

mattwthompson:
#249 is merged and this should be good to go stand-alone, although there is now some cross-talk between this and #252

PR #257, opened 2022-04-28
[Evaluator] Add option to restart failed calculations

SimonBoothroyd:
## Description

This PR adds the ability for the Evaluator target to restart any failed calculations up to a specified number of times before failing. This adds some resilience in cases of rare failures due to e.g. calculations landing on a badly configured compute node.


leeping:
@j-wags This is totally on me for failing to review it after several months. Is this code still correct? Do we still want this in a future release?

PR #262, opened 2022-08-01
Remove support for deprecated `simtk` namespace

mattwthompson:
This PR removes soft checks for the `simtk` namespace, which was deprecated last fall and will presumably be removed at some point. This decreases the likelihood that errant warnings will be printed out to the user.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,8 +22,6 @@ jobs:
           - macOS-latest
           - ubuntu-latest
         python-version:
-          - "3.6"
-          - "3.7"
           - "3.8"
           - "3.9"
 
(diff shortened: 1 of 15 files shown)


leeping:
A number of students in my group currently need to use OpenMM <7.5 in their environments because we are getting some difficult-to-decipher linking errors when linking OpenMM >= 7.5 to TeraChem for QM/MM calculations.  The "safe" version that we are currently using is 7.4.1.  Could we hold off on merging this PR until the issues are definitively resolved on our end?

mattwthompson:
Sure!

leeping:
Thanks - really appreciate it. :) We can definitely open it later.

PR #263, opened 2022-08-09
Update global parameters in a `CustomBondForce`

jthorton:
A `CustomBondForce` is commonly used to implement scaled14 interactions excluded from a `CustomNonbondedForce` and can contain different global parameters, for example, the scale parameter to scale the interactions by. This PR also checks the custom bond force for global parameters when copying forces to a new openmm system. See this issue https://github.com/openforcefield/smirnoff-plugins/issues/36


leeping:
This looks good to me. Sorry for the extremely late review. I'll merge it now!

Issue #260, opened 2022-07-28
Pymbar 4 compatibility 


jthorton:
The new release of [pymbar](https://github.com/choderalab/pymbar/releases/tag/4.0.1) has some API changes which have broken the imports in Forcebalance. Maybe it would be good to add a nightly CI test to forcebalance to catch this early?

pavankum:
I think @mattwthompson is working on it in #259.

mattwthompson:
I started in on it the other day, kinda grouping a number of API breaks together. Hope to have everything finished up early next week, can split it out into smaller bits if anything is really timely.

> Maybe it would be good to add a nightly CI test to forcebalance to catch this early?

Agree

mattwthompson:
I updated the API in https://github.com/leeping/forcebalance/pull/261 but there's no clear path forward. The feedstock was updated a while back with this constraint in mind (https://github.com/conda-forge/forcebalance-feedstock/pull/23/files#diff-f3725a55bf339595bf865fec73bda8ac99f283b0810c205442021f29c06eea9a). Some other software might be compatible with both versions 3 and 4.

PR #259, opened 2022-07-25
OpenFF Toolkit v0.11.0+ compatibility

mattwthompson:
Closes #266, #264, #261, #260

PR #267, opened 2022-10-26
[WIP] OpenFF Toolkit 0.11 compatibility (without pymbar update)

j-wags:
This is an attempt to more cleanly separate the OFFTK 0.11 changes (#259) from the pymbar update PR (#260). I think #259 didn't run the full test suite because it came from a fork. So if this works, this PR should supersede #259.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,8 +22,6 @@ jobs:
           - macOS-latest
           - ubuntu-latest
         python-version:
-          - "3.6"
-          - "3.7"
           - "3.8"
           - "3.9"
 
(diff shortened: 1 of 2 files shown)


j-wags:
Hm, I think I'll need to run these tests locally, since the ones that look relevant are getting skipped - Specifically I'd expect test_system [TestEvaluatorBromineStudy](https://github.com/leeping/forcebalance/blob/78a48d0e9ee9b5a8f3295cee8a811ad4cfcd2aa6/src/tests/test_system.py#L182) and [TestOpenFFTorsionProfileStudy](https://github.com/leeping/forcebalance/blob/78a48d0e9ee9b5a8f3295cee8a811ad4cfcd2aa6/src/tests/test_system.py#L249) to be useful in determining compatibility with the 0.11 toolkit. 

I'm off until Monday but will start here when I return.

PR #272, opened 2022-11-17
Fix coverage

mattwthompson:
* [ ] Fix path to collecting coverage data
* [ ] Ensure coverage reports are uploaded


mattwthompson:
Depends on #271 and #269

mattwthompson:
These reports are wrong; I can't figure out a way to properly set up the path discovery with the `src/` layout

PR #285, opened 2023-06-27
Remove validate cache since the ChemicalEnvironment class is being dropped in OFFTK 0.14

### LINKED ISSUE #286: Unexpected keyword argument 'return_topology'
As of [this commit](https://github.com/openforcefield/openff-toolkit/commit/3f864d5598bc25096198955fc9ed629064deb311), `openff-toolkit.typing.engines.smirnoff.Forcefield.create_openmm_system` no longer takes `return_topology` as a keyword argument, so it throws an error on openff-toolkit version 0.14.0 with forcebalance version 1.9.5. To quote from the breaking changes listed in that commit

> To access the `Topology` that results from parametrization, call `ForceField.create_interchange` and access the `.topology` attribute of the returned `Interchange` object.

Although the [documentation for that method](https://docs.openforcefield.org/projects/toolkit/en/stable/api/generated/openff.toolkit.typing.engines.smirnoff.ForceField.html#openff.toolkit.typing.engines.smirnoff.ForceField.create_interchange) says that API is experimental.

It sounds like a straightforward fix, so I can submit a PR for it if you want.

j-wags:
ForceBalance has a [cache for the `ChemicalEnvironment.validate` method](https://github.com/leeping/forcebalance/blob/10b9280cab9735634d25a4f9c8b3e84cb13a6bb1/src/smirnoff_hack.py#L75C1-L84). This hasn't been used for a while (we stopped validating SMARTS a while ago and now just treat them as strings instead of ChemicalEnvironments), and we're finally dropping the ChemicalEnvironment class in the 0.14.0 OpenFF Toolkit release. It should suffice to simply remove the cache here. 

- [x] Remove `ChemicalEnvironment.validate` cache
- [x] Fixes #286 by replacing usage of `return_topology` with Interchange equivalents

CI on GitHub seems to be broken, but I tested the changes locally and was able to run study 22 successfully.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,10 +91,14 @@ jobs:
         cd gromacs-5.1.5/cmake
         cmake -DCMAKE_INSTALL_PREFIX=$GITHUB_WORKSPACE/opt/gromacs/5.1.5 -DGMX_DOUBLE=ON -DGMX_BUILD_OWN_FFTW=ON ..
         make -j 4 && make install
+        cd ..
+        mkdir cmake_s
+        cd cmake_s
+        cmake -DCMAKE_INSTALL_PREFIX=$GITHUB_WORKSPACE/opt/gromacs/5.1.5 -DGMX_BUILD_OWN_FFTW=ON ..
+        make -j 4 && make install
+        cd ..
         . $GITHUB_WORKSPACE/opt/gromacs/5.1.5/bin/GMXRC.bash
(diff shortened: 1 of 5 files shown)


j-wags:
Apologies - It looks like there was another change that affected ForceBalance (`create_openmm_system` had the `return_topology` kwarg removed). I've updated the PR to fix that as well.

leeping:
I have some time to work on the next release, so I'll merge this PR as I continue to work on it.

j-wags:
Great - Thanks, Lee Ping!

PR #292, opened 2023-10-13
Permit fits with virtual sites

mattwthompson:


mattwthompson:
It runs @lilyminium's MRE without error, I don't know if the results are correct or not

leeping:
I'd like to have a minimal test case if possible, for example, we could fit the energies and forces on a water dimer (we could even use a force field to generate fake QM data.)  If we start the optimization from different parameters (p1) than those that were used to generate the fake data (p0), and the optimization converges to p0, then we will know the results are correct.

mattwthompson:
Here's the test case that does not work with 1.9.6: [vsites-interchange.tar.gz](https://github.com/leeping/forcebalance/files/12921504/vsites-interchange.tar.gz)

What tests currently in the codebase should I model a new set of tests after?

leeping:
I think this would work well: https://github.com/leeping/forcebalance/blob/master/src/tests/test_system.py#L122

The test involves running the optimization in `studies/001_water_tutorial`, which consists of fitting energies and forces for a number of water clusters (I believe the data set includes a few hundred structures 6-mers and 12-mers).  To make the current test more lightweight, I think it should be enough to use a few dozen structures of dimers.

mattwthompson:
That's a pretty heavy test, is there a way to test this behavior with fewer moving parts and files/configuration (akin to a set of unit tests)?

leeping:
You could create an instance of the ForceBalance SMIRNOFF engine from a given .mol2 and force field XML file, calculate the single point energy and force for the structure, and compare it to a reference value.  

A similar example for AMBER / GROMACS / TINKER is here:

https://github.com/leeping/forcebalance/blob/master/src/tests/test_engine.py#L116

PR #293, opened 2023-12-05
Trim un-used dependencies

mattwthompson:
There are a couple of things brought in that are no longer needed - trimming them out of the test environment will make it run (even) faster


devtools/conda-envs/test_env.yaml
```diff
@@ -10,7 +9,6 @@ dependencies:
     # Testing
   - pytest
   - pytest-cov
-  - codecov
```
mattwthompson, line 13: This is no longer needed on the user side - the official action handles uploading itself

PR #294, opened 2023-12-05
Use double-precision GROMACS builds from `conda-forge`

mattwthompson:
Some context: https://github.com/conda-forge/gromacs-feedstock/blob/cde675cda173295466795a1e8d69940d4343723e/recipe/meta.yaml#L19-L24


mattwthompson:
This succeeds in pulling down builds (instead of building new ones on the fly) but more work is needed to update the tests to use modern versions of GROMACS

PR #295, opened 2024-01-18
Do not use `numpy.asfarray` which is removed in 2.0

mattwthompson:
https://numpy.org/devdocs/release/2.0.0-notes.html#numpy-2-0-python-api-removals

```
$ git checkout upstream/master && ruff check --preview --select NPY201 . --exclude tools
HEAD is now at b395fd4b Replace &> run.log with > run.log 2>&1 (dash compatible)
src/molecule.py:3173:22: NPY201 `np.asfarray` will be removed in NumPy 2.0. Use `np.asarray` with a `float` dtype instead.
Found 1 error.
```

### DIFF
--- a/src/molecule.py
+++ b/src/molecule.py
@@ -3170,7 +3170,7 @@ class Molecule(object):
             elif result == -1:
                 break
             #npa    = np.array(xyzvec)
-            xyz    = np.asfarray(xyzvec)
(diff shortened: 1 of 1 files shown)


leeping:
Thank you, Matt - I'll merge it now.  I think the test failures are unrelated.

leeping (approved):

PR #302, opened 2024-08-20
Fix pymbar import

abidart:
Newest versions of pymbar are imported this way


mattwthompson:
There's much more work involved with upgrading this one https://github.com/leeping/forcebalance/pull/261

PR #311, opened 2026-02-25
Update for modern openff and get CI working again

lilyminium:
This supersedes #310 and updates CI to use modern environments and modern OpenFF packages.

**Changes**
- removes pkg_resources
- fixes gromacs install (required modifications for CMAKE versions and downloading fftw)
- fixes the Evaluator bromine study targets. The units for minimization were incompatible with OpenMM; at some point the tolerance switched from kcal/mol to kcal/mol/nm.
- Allows all tests to run instead of stopping at the first error

This passes CI on my fork. https://github.com/lilyminium/forcebalance/pull/1

PR #312, opened 2026-02-26
Expand CI to Python 3.9

lilyminium:
This PR updates the test matrix to go back to Python 3.9. It was unexpected much harder than I thought would be required, hence the number of commits -- I'd recommend squash merging if this is merged!

**Changes made**
- I've added a new env file for Python 3.9. The main difficulty I ran into is that ambertools is not available on Python 3.9 which short-circuited a lot of OpenFF tests. I dealt with this by just not running OpenFF tests on Python 3.9
- I think I fixed the ultimate cause of the flaky Evaluator bug, fingers crossed. I had thought this was fixed in #311 but the green CI there may have just been a fluke. I've commented where the additional code is for Evaluator.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,8 @@ jobs:
     - name: Set up conda environment
       uses: mamba-org/setup-micromamba@v1
       with:
-        environment-file: devtools/conda-envs/test_env.yaml
+        # ambertools has no Python 3.9 builds; use a separate env file for 3.9
+        environment-file: ${{ matrix.python-version == '3.9' && 'devtools/conda-envs/test_env_py39.yaml' || 'devtools/conda-envs/test_env.yaml' }}
         create-args: >- # beware the >- instead of |, we don't split on newlines but on spaces
           python=${{ matrix.python-version }}
 
@@ -49,12 +50,6 @@ jobs:
       if: matrix.python-version == '3.9' || matrix.python-version == '3.10'
       shell: bash -l {0}
       run: pip install "setuptools<82"
-
-    - name: Install ambertools for Python > 3.9
-      # AmberTools is not available for Python 3.9, so we install it separately for Python 3.10 and above
-      if: matrix.python-version != '3.9'
-      shell: bash -l {0}
-      run: micromamba install -c conda-forge ambertools
           
     - name: Additional info about the build
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,8 @@ jobs:
           - macOS-latest
           - ubuntu-latest
         python-version:
+          - "3.9"
+          - "3.10"
           - "3.11"
           - "3.12"
           # pymbar 3 does not support Python 3.13
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
       # AmberTools is not available for Python 3.9, so we install it separately for Python 3.10 and above
       if: matrix.python-version != '3.9'
       shell: bash -l {0}
-      run: mamba install -c conda-forge ambertools
+      run: micromamba install -c conda-forge ambertools
           
     - name: Additional info about the build
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,6 +44,11 @@ jobs:
         environment-file: devtools/conda-envs/test_env.yaml
         create-args: >- # beware the >- instead of |, we don't split on newlines but on spaces
           python=${{ matrix.python-version }}
+
+    - name: Pin setuptools for Python < 3.10
+      if: matrix.python-version == '3.9' || matrix.python-version == '3.10'
+      shell: bash -l {0}
+      run: pip install "setuptools<82"
           
     - name: Additional info about the build
       shell: bash
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,12 @@ jobs:
       if: matrix.python-version == '3.9' || matrix.python-version == '3.10'
       shell: bash -l {0}
       run: pip install "setuptools<82"
+
+    - name: Install ambertools for Python > 3.9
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -39,9 +41,15 @@ jobs:
...
-        environment-file: devtools/conda-envs/test_env.yaml
+        # ambertools has no Python 3.9 builds; use a separate env file for 3.9
+        environment-file: ${{ matrix.python-version == '3.9' && 'devtools/conda-envs/test_env_py39.yaml' || 'devtools/conda-envs/test_env.yaml' }}
         create-args: >- # beware the >- instead of |, we don't split on newlines but on spaces
           python=${{ matrix.python-version }}
+
+    - name: Pin setuptools for Python < 3.10
+      if: matrix.python-version == '3.9' || matrix.python-version == '3.10'
+      shell: bash -l {0}
+      run: pip install "setuptools<82"
```
lilyminium, line 52: This is to deal with the pkg_resources bug. It's been removed in the newer versions of openff dependencies, so I chose to just pin setuptools for older versions of Python here.

devtools/conda-envs/test_env.yaml
```diff
@@ -24,6 +24,6 @@ dependencies:
   - geometric
   # - gromacs =2019.1
   - openff-toolkit >=0.11.3
-  - openff-evaluator >= 0.4.1
+  - openff-evaluator-base >= 0.4.5
```
lilyminium, line 27: I think this may have been the ultimate cause of the flaky Evaluator test.

src/tests/test_system.py
```diff
@@ -12,6 +13,11 @@
 import numpy as np
 import pytest
 
+skip_openff_py39 = pytest.mark.skipif(
+    sys.version_info < (3, 10),
+    reason="openff packages require ambertools which requires Python >= 3.10",
+)
+
```
lilyminium, line 20: I just skip all OpenFF tests in Python 3.9 because it's too hard to resolve a working environment

src/tests/test_system.py
```diff
@@ -192,25 +200,67 @@ def setup_method(self, method):
...
+                if ready_marker in f.read():
+                    break
+            time.sleep(0.5)
+        else:
+            self.estimator_process.terminate()
+            log_contents = open(self._server_log_path).read()
+            pytest.fail(
+                "Evaluator server did not start within 120 seconds. Log:\n%s"
+                % log_contents[-2000:]
+            )
```
lilyminium, line 238: I added this to help debug failing Evaluator tests, would be great to keep it in for future debugging too.

src/tests/test_system.py
```diff
@@ -192,25 +200,67 @@ def setup_method(self, method):
...
     def teardown_method(self):
         self.estimator_process.terminate()
-        shutil.rmtree("working_directory")
-        shutil.rmtree("stored_data")
+        self._server_log.close()
+        if os.path.exists(self._server_log_path):
+            os.remove(self._server_log_path)
+        for dnm in ["working_directory", "stored_data"]:
+            if os.path.exists(dnm):
+                shutil.rmtree(dnm)
```
lilyminium, line 249: I added a lot of catches for removing working directories because it can cause additional errors if a test exits early.

src/tests/test_system.py
```diff
@@ -192,25 +200,67 @@ def setup_method(self, method):
...
-        data      = objective.Full(np.zeros(objective.FF.np),1,verbose=True)
+        try:
+            data = objective.Full(np.zeros(objective.FF.np),1,verbose=True)
+        except Exception as exc:
+            self._server_log.flush()
+            log_contents = open(self._server_log_path).read()
+            raise RuntimeError(
+                "objective.Full raised %s: %s\nServer log (last 2000 chars):\n%s"
+                % (type(exc).__name__, exc, log_contents[-2000:])
+            ) from exc
```
lilyminium, line 263: Part of the same diagnostics for Evaluator above

studies/003d_evaluator_liquid_bromine/run_server.py
```diff
@@ -1,7 +1,25 @@
 #!/usr/bin/env python3
+import multiprocessing
+
+# Use "spawn" instead of the Linux default "fork". When forking from the
+# multi-threaded Dask/Tornado process, child processes inherit locked mutexes
+# from sibling threads and deadlock immediately. "spawn" starts a fresh
+# interpreter with no inherited thread state, avoiding the deadlock entirely.
+# This is the primary fix for Python 3.9 / older openff-evaluator versions.
+multiprocessing.set_start_method("spawn", force=True)
+
```
lilyminium, line 11: Claude claimed this would help the Evaluator test. I think ultimately the cause was actually there was a version mismatch between Evaluator and the Toolkit, but it doesn't *hurt*.

.gitignore
```diff
@@ -123,3 +123,12 @@ studies/*/targets/
 studies/*/*.tmp/
 studies/*/*.bak/
 sutdies/*/*.err
+
+# test files from running CI locally
```
lilyminium, line 127: Just to avoid accidental commits

src/tests/test_system.py
leeping: That makes sense, especially if OpenFF does not support Python 3.9.

src/tests/test_system.py
leeping: Thanks, this looks useful.

src/tests/test_system.py
leeping: I don't remember if rmtree threw an error if the file / folder doesn't exist, but yes, this makes sense.

studies/003d_evaluator_liquid_bromine/run_server.py
leeping: Interesting, could you add that Claude suggested it, since I don't fully understand the comment either?  That would be helpful if it causes issues later on.

.gitignore
leeping: src/tests/files is a global folder for storing data files used by the tests, but these filenames makes it look like some temp-files were generated when you ran tests locally.  I don't remember if this happened when I ran tests locally.  Either way it should be fine.

leeping (approved):

PR #315, opened 2026-03-05
Fix vsite fits

lilyminium:
This PR enables ForceBalance to optimize virtual site (v-site) parameters in SMIRNOFF force fields. Previously, the code raised a hard NotImplementedError for any SMIRNOFF system with virtual sites. This PR removes that block and implements the full pipeline needed for v-site parameter fitting. As usual I've tried to add explanatory comments to code changes.

This is rebased off of #312 so the earlier commits aren't relevant -- as usual would recommend squash-merging over merging this though!

**Biggest changes**

- I added some new test files, these are the primary culprit for the 200k added lines. I can compress the targets if that's easier.
- added a select_virtual_site_parameter to get virtual site parameters from OpenFF infrastructure
- undid the commenting of unit_str so we can use parameter_eval and get parameters with units
- fixed pgrad updating by scanning virtual site keys properly and getting SMIRKS properly
- allowed pickling of Versions with protocol 0, this was recently erroring due to packaging version 26.0
- Changed the behaviour of Evaluator_SMIRNOFF to not block on target submission, but target retrieval instead -- otherwise every other target has to wait until Evlauator is done before running.

**New tests**
I added two tests:
1. studies/028: testing whether co-optimizing water and another molecule vdW would execute without erroring when fit pure water Liquid properties, AbInitio clusters, and Evaluator mixture properties. Unfortunately this is actually very slow, ~20 min to execute.
2. studies/029: testing agreement of OpenMM and OpenFF gradients for TIP4P and TIP5P


src/smirnoffio.py
```diff
@@ -88,8 +181,42 @@ def smirnoff_analyze_parameter_coverage(forcefield, tgt_opts):
             optgeo_options_txt = os.path.join(target_path, tgt_option['optgeo_options_txt'])
             sys_opts = forcebalance.opt_geo_target.OptGeoTarget.parse_optgeo_options(optgeo_options_txt)
             mol2_paths = [os.path.join(target_path,fnm) for sysopt in sys_opts.values() for fnm in sysopt['mol2']]
+        elif tgt_option['type'] == 'EVALUATOR_SMIRNOFF':
```
lilyminium, line 196: Previously Evaluator data was not part of smirnoff_parameter_coverage. This is very optional but nice to have to track when we've done something wrong.

lilyminium:
> 2073.84s call     src/tests/test_system.py::TestEvaluatorWaterVSiteStudy::test_water_vsite_study

Ohhhh, that's not good.

lilyminium:
It's now down to 1300 seconds, which is still quite long -- maybe this test should be skipped for speed reasons.

src/tests/test_smirnoffio.py
```diff
@@ -0,0 +1,291 @@
+import pytest
+
+has_openff_toolkit = True
```
lilyminium, line 3: This entire file just tests the new vsite parsing and assignment logic.

src/tests/test_system.py
```diff
@@ -181,41 +207,112 @@ def test_thermo_bromine_study(self):
         """Check liquid bromine study (Thermo target) converges to expected results"""
         self.run_optimizer()
 
-class TestEvaluatorBromineStudy(ForceBalanceSystemTest):
+
+class EvaluatorServerMixin:
```
lilyminium, line 208: Refactor the Evaluator tests to not repeat so much

src/forcefield.py
```diff
@@ -747,7 +747,7 @@ def addff_xml(self, ffname):
                 res = re.search(r'^[-+]?[0-9]*\.?[0-9]*([eEdD][-+]?[0-9]+)?', quantity_str)
                 value_str, unit_str = quantity_str[:res.end()], quantity_str[res.end():]
                 # LPW 2023-01-23: Behavior of parameter unit string for "evaluated" parameter is undefined.
-                unit_str = ""
+                # unit_str = ""
```
lilyminium, line 750: Fixes parameter_eval

src/openmmio.py
```diff
@@ -270,6 +270,11 @@ def GetVirtualSiteParameters(system):
                 vsprm.append(_openmm.OutOfPlaneSite_getWeight12(vs))
                 vsprm.append(_openmm.OutOfPlaneSite_getWeight13(vs))
                 vsprm.append(_openmm.OutOfPlaneSite_getWeightCross(vs))
+            elif vstype == 'LocalCoordinatesSite':
```
lilyminium, line 273: OpenFF vsites are convereted to localcoordinatesites

devtools/conda-envs/test_env.yaml
```diff
@@ -23,7 +23,28 @@ dependencies:
...
+  - requests
+  - python-dateutil
+  - pydantic =2,<2.12
+  - taproom
+  - dataclasses
+  - pandas =2
+
+  - pip:
+    # install branch of evaluator
+    - git+https://github.com/openforcefield/openff-evaluator.git@fix-vsite-fits
```
lilyminium: We will aim to get this merged and into a release ASAP, but this branch needed to be developed in tandem with the Evaluator one for testing to work (hence the same name)

src/evaluator_io.py
```diff
@@ -525,9 +560,7 @@ def submit_jobs(self, mvals, AGrad=True, AHess=True):
         )
 
         if (
-            self._pending_estimate_request.results(
-                True, polling_interval=self._options.polling_interval
-            )[0] is None
+            self._pending_estimate_request.results(False)[0] is None
```
lilyminium, line 563: Having this blocking request here originally meant that other targets had to wait until Evaluator was done computing to even start -- which can be a while ...

src/evaluator_io.py
```diff
@@ -738,6 +771,19 @@ def get(self, mvals, AGrad=True, AHess=True):
...
+        estimation_results, _ = self._pending_estimate_request.results(
+            True, polling_interval=self._options.polling_interval
+        )
+        if estimation_results is None:
+            raise RuntimeError(
+                "No `EvaluatorServer` could be found to retrieve results from. "
+                "Please double check that a server is running, and that the connection "
+                "settings specified in the input script are correct."
+            )
+
```
lilyminium, line 786: I moved (and added the same runtimeerror check) this from the submit_jobs function above.

leeping (approved):

PR #316, opened 2026-03-10
Add pairwise abinitio target

lilyminium:
This is the target @chapincavender has been using in his protein force field fits. It's only supported for the SMIRNOFF engine at the moment. I've largely completely ported over the code in Chapin's branch at https://github.com/chapincavender/forcebalance/tree/pairwise , which he's been using in fits for multiple years now. The only changes I've made are very minor updates (importing `plot_qm_vs_mm`, erroring early on force fitting) and I added a minimal test.

Python 3.10 is erroring because it can't resolve an environment with the openff-toolkit>0.18 -- if #315 is merged and this is rebased off that, we will skip testing on Python 3.9 and 3.10 and only tests on Pythons 3.11-12.


leeping (approved):

leeping:
Thanks a lot!  This PR looks nice and clean so I'll go ahead and merge. :)

PR #317, opened 2026-04-01
Update pymbar

lilyminium:
This PR adds a test to compare pymbar v3 and v4 compatiblity and also updates for v4. The main change seems to be moving from the default 'adaptive' solver in v3 to the 'robust' protocol in v4, which also calls the adaptive solver method.


src/tests/test_liquid.py
```diff
@@ -2,13 +2,63 @@
...
+FIXTURE_DIR = os.path.join(os.path.dirname(__file__), 'files', 'test_liquid')
+
+# U_kln.npy — shape (6, 6, 801), float64
+#   Reduced potential energy matrix U_kln[k, m, n] = (E_k[n] + P_m * V_k[n] * pvkj) * beta_m
+#   where k = source simulation index, m = evaluation state index, n = snapshot index.
+#   Built from the six npt_result.p pickles in files/test_liquid/single.tmp/Liquid/iter_0000/
+#   (water at 249.15 K/1 atm, 273.15 K/1 atm, 298.15 K/1 atm, 373.15 K/1 atm,
+#    298.15 K/20 bar, 298.15 K/2000 bar; 801 snapshots each).
+#   Regenerate with: conda run -n <env> python tools/mbar_mre.py --save-ref
+U_KLN_PATH  = os.path.join(FIXTURE_DIR, 'U_kln.npy')
```
lilyminium, line 23: From here downwards: added tests comparing direct pymbar execution. The script referenced here is attached, not sure if I should add it to the repo or if it's clutter.

[mbar_mre.py](https://github.com/user-attachments/files/26422225/mbar_mre.py)

src/liquid.py
```diff
@@ -24,7 +24,18 @@
 try:
     from lxml import etree
 except: pass
-from pymbar import pymbar
+try:
+    from pymbar import pymbar  # pymbar 3: MBAR lives in pymbar.pymbar submodule
+    _MBAR_SOLVER_KW = {}  # v3: self-consistent-iteration default is fine
+except ImportError:
+    import pymbar  # pymbar 4: MBAR lives at pymbar top-level
+    _MBAR_SOLVER_KW = {'solver_protocol': 'robust'}  # v4: default hybr diverges on some data
```
lilyminium, line 32: v3: used the adaptive solver by default. (https://pymbar.readthedocs.io/en/3.1.1/mbar.html#pymbar.MBAR)
v4: 'robust' uses adaptive, followed by L-BFGS (https://deepwiki.com/choderalab/pymbar/2.1-mbar-implementation) . 

I'm not sure why, but plain swapping this out for the plain `adaptive` protocol results in tests failing, so maybe the fallback to L-BFGS is necessary. @mattwthompson I saw you managed to swap it out for `adaptive` in your PR, did you notice any differences?

mattwthompson: Unfortunately I don't recall, I was probably trying each option until one worked without really understanding the underlying behavior

leeping:
Thanks so much for adding the pymbar tests.  I agree this is ready to merge.

leeping (approved):

PR #321, opened 2026-06-04
Fix recharge compatibility

lilyminium:
This PR fixes recharge compatibility.

Note -- recharge_io.py only works for BCC's (and has only ever worked for them) -- it does not work for training virtual sites (or any FF that has vsites with charges, even if they're not trained). I added an error for that case.

The main changes are:
- I think the new targets.tar.gz in `master` may have been generated with numpy v2, either way I needed to regenerate for v1 compatibility
- updating to new API, biggest difference is that recharge now expects reference values instead of target residuals, but overall the diff is pretty small.
- also added some explicit comments and guards about how this will not work for virtual sites (which it never did)

CI is failing for a couple reasons:
- TINKER is being odd [?] -- will look into this
- the flaky bromine test a couple times


leeping (approved):

leeping:
Thank you, Lily and Matt!

